PR #1, opened 2022-07-12
feat: inventory operator

troian:
ref ovrclk/engineering#360


operator/cmd.go
```diff
@@ -0,0 +1,19 @@
+package operator
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/ovrclk/provider-services/operator/inventory"
+)
+
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
```
tidrolpolelsef, line 10: Declare the Args this command takes (or doesn't take) ?

operator/inventory/ceph.go
```diff
@@ -0,0 +1,325 @@
...
+
+	clusters := make(cephClusters)
+	scs := make(cephStorageClasses)
+
+	scrapeData := resp{
+		res: nil,
+		err: cephInventoryInProgress,
+	}
+
+	var scrapech <-chan runner.Result
```
tidrolpolelsef: Is there a reason why this can't be done with `scrapech := ` ?

operator/inventory/cmd.go
```diff
@@ -0,0 +1,292 @@
...
+				if inventory, valid := res.Value().([]akashv2beta1.InventoryClusterStorage); valid {
+					inv.Spec.Storage = append(inv.Spec.Storage, inventory...)
+				}
+			}
+		}
+
+		var err error
+		if data, err = json.Marshal(&inv); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			data = []byte(err.Error())
```
tidrolpolelsef, line 268: What does assigning `data =` do here? Maybe I am just not seeing what the effect of that is.

tidrolpolelsef: OK, maybe this works with the `defer` above? But couldn't `w.writeHeader` get called twice? Also It's kind of an odd pattern.

operator/inventory/types.go
```diff
@@ -0,0 +1,126 @@
...
+	CtxKeyInformersFactory = ContextKey("informers-factory")
+)
+
+type resp struct {
+	res []akashv2beta1.InventoryClusterStorage
+	err error
+}
+
+type req struct {
+	resp chan resp
```
tidrolpolelsef: So this member is defined as type `resp`, but is actually a channel which has values in it of type `resp`? It would probably be better to call this `respCh`, I think.

troian: 💯

operator/inventory/cmd.go
troian: if `w.WriteHeader` is not called explicitly the call to `w.Write` triggers implicit `WriteHeader(http.StatusOK)`

operator/inventory/ceph.go
troian: 💯

PR #2, opened 2022-07-13
refactor: set default value for kubeconfig flag

troian:
expand env if any for kubeconfig path

ref ovrclk/engineering#360

### DIFF
--- a/cluster/kube/clientcommon/open_kube_config.go
+++ b/cluster/kube/clientcommon/open_kube_config.go
@@ -2,32 +2,32 @@ package clientcommon
 
 import (
 	"fmt"
+	"os"
+
 	"github.com/tendermint/tendermint/libs/log"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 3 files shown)


cluster/kube/clientcommon/open_kube_config.go
```diff
@@ -2,19 +2,18 @@ package clientcommon
...
-	"k8s.io/client-go/util/homedir"
-	"os"
-	"path"
 )
 
 func OpenKubeConfig(cfgPath string, log log.Logger) (*rest.Config, error) {
 	// If no value is specified, use a default
-	if len(cfgPath) == 0 {
-		cfgPath = path.Join(homedir.HomeDir(), ".kube", "config")
+	if cfgPath != "" {
```
tidrolpolelsef: I'm a little confused by this, it used to be if the **string was empty**, the use a default of `$HOME/.kube/config` (in a round about way)

Now, it's going to expand the string **if nonempty** ?

troian: updated with comment and changed logic a bit

PR #50, opened 2022-11-23
fix(provider): restore the command field in the manifest

troian:
fixes ovrclk/engineering#647

### DIFF
--- a/pkg/apis/akash.network/crd.yaml
+++ b/pkg/apis/akash.network/crd.yaml
@@ -50,6 +50,10 @@ spec:
                             type: string
                           image:
                             type: string
+                          command:
+                            type: array
+                            items:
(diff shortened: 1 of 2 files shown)


andy108369:
For patching the CRD when upgrading provider from `0.1.0` to `0.1.1`:

```
kubectl patch crd manifests.akash.network --type='json' -p='[{"op": "add", "path": "/spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/group/properties/services/items/properties/command", "value":{"type":"array"}}, {"op": "add", "path": "/spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/group/properties/services/items/properties/command/items", "value":{"type":"string"}}]'
```

PR #51, opened 2022-11-24
fix: fail provider start if lease namespace is not found

troian:
fixes ovrclk/engineering#463


boz (approved):
:rocket:

PR #57, opened 2022-12-05
Set LeaseIPStatus json tags to keep consistency

cloud-j-luna:
Set the json tags for LeaseIPStatus type.
Is there any reason to not use the net.IP type for the IP?

### DIFF
--- a/gateway/rest/types.go
+++ b/gateway/rest/types.go
@@ -5,10 +5,10 @@ import (
 )
 
 type LeasedIPStatus struct {
-	Port         uint32
-	ExternalPort uint32
-	Protocol     string
-	IP           string
+	Port         uint32 `json:"port"`
+	ExternalPort uint32 `json:"external_port"`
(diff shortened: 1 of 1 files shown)


gateway/rest/types.go
```diff
@@ -5,10 +5,10 @@ import (
 )
 
 type LeasedIPStatus struct {
-	Port         uint32
-	ExternalPort uint32
-	Protocol     string
-	IP           string
+	Port         uint32 `json:"port"`
+	ExternalPort uint32 `json:"external_port"`
```
troian, line 9: this is potentially breaking change as the default values are different
https://go.dev/play/p/nh3PaU5Qatb

cloud-j-luna: These are breaking changes indeed, but they make the whole object consistent in terms of casing which is an overall improvement and might be worth to include in a future release.

PR #82, opened 2023-01-31
Fixed typo in provider image url (removed extra colon)

jigar-arc10:


jigar-arc10:
It looks like this is also not working, as none of the packages are tagged as "stable".

troian:
the fix is correct, thank you, however, those files are only used in testing and running `_run` examples and are not targeted for production.

jigar-arc10:
Yes, Helm is not using this in production, but Praetor uses this as when Praetor started, Helm was on the beta version. We will be updating our process soon to helm, but in the meantime, we wanted to have this fixed. Also, Can you label v0.2.1 to stable as well?

troian:
interesting. the best I would recommend for Praetor is to pull those files somewhere and affix for whatever values are suitable.

those files are always for developers only and can be broken at any time.

also `stable` is meant for the `node` repo only. the provider does not use it.

troian:
sorry, let me recap.
the `stable` tag is used with the provider image built locally or in CI actions, so in _docs/kustomize the image must remain tagged `stable`. if you need those files to deploy, copy them somewhere and tag with version that is suitable for Praetor

jigar-arc10:
Until we move to the helm, I have forked the provider and changed appropriate files and will make sure to have up to date with Akash-network/provider

troian:
lets set the image on all files in the PR to `ghcr.io/akash-network/provider:stable`

PR #86, opened 2023-02-15
feat: allow cert-manager annotations on ingress based on environment variables

cloud-j-luna:
This PR allows cert-manager annotations on ingress based on environment variables.

### DIFF
--- a/cluster/kube/client.go
+++ b/cluster/kube/client.go
@@ -3,6 +3,7 @@ package kube
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -56,6 +57,7 @@ type client struct {
 	ns                string
 	log               log.Logger
 	kubeContentConfig *restclient.Config
+	env               map[string]string
 }
 
 func (c *client) String() string {
@@ -98,6 +100,7 @@ func NewClient(ctx context.Context, log log.Logger, ns string, configPath string
 		ns:                ns,
 		log:               log.With("client", "kube"),
 		kubeContentConfig: config,
+		env:               environmentVariablesToMap(),
 	}, nil
 }
 
@@ -766,4 +769,15 @@ func (c *client) deploymentsForLease(ctx context.Context, lid mtypes.LeaseID) (m
 
 func (c *client) KubeVersion() (*version.Info, error) {
 	return c.kc.Discovery().ServerVersion()
+}
+
+func environmentVariablesToMap() map[string]string {
+	m := make(map[string]string)
+	for _, e := range os.Environ() {
+		if i := strings.Index(e, "="); i >= 0 {
+			m[e[:i]] = e[i+1:]
+		}
+	}
+
+	return m
 }
--- a/cluster/kube/client.go
+++ b/cluster/kube/client.go
@@ -3,7 +3,7 @@ package kube
 import (
 	"context"
 	"fmt"
-	"os"
+	"github.com/akash-network/provider/cluster/util"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -100,7 +100,7 @@ func NewClient(ctx context.Context, log log.Logger, ns string, configPath string
 		ns:                ns,
 		log:               log.With("client", "kube"),
 		kubeContentConfig: config,
-		env:               environmentVariablesToMap(),
+		env:               util.EnvironmentVariablesToMap(),
 	}, nil
 }
 
@@ -769,15 +769,4 @@ func (c *client) deploymentsForLease(ctx context.Context, lid mtypes.LeaseID) (m
 
 func (c *client) KubeVersion() (*version.Info, error) {
 	return c.kc.Discovery().ServerVersion()
-}
-
-func environmentVariablesToMap() map[string]string {
-	m := make(map[string]string)
-	for _, e := range os.Environ() {
-		if i := strings.Index(e, "="); i >= 0 {
-			m[e[:i]] = e[i+1:]
-		}
-	}
-
-	return m
 }
--- a/cluster/kube/client_ingress.go
+++ b/cluster/kube/client_ingress.go
@@ -26,7 +26,7 @@ const (
 	akashIngressClassName = "akash-ingress-class"
 )
 
-func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective) map[string]string {
+func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective, env map[string]string) map[string]string {
 	// For kubernetes/ingress-nginx
 	// https://github.com/kubernetes/ingress-nginx
 	const root = "nginx.ingress.kubernetes.io"
@@ -66,6 +66,12 @@ func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDir
 		}
 	}
 
+	if env["AKASH_PROVIDER_ISSUER_TYPE"] == "cluster-issuer" {
+		result["cert-manager.io/cluster-issuer"] = env["AKASH_PROVIDER_ISSUER_NAME"]
+	} else if env["AKASH_PROVIDER_ISSUER_TYPE"] == "issuer" {
+		result["cert-manager.io/issuer"] = env["AKASH_PROVIDER_ISSUER_NAME"]
+	}
+
 	result[fmt.Sprintf("%s/proxy-next-upstream", root)] = strBuilder.String()
 	return result
 }
@@ -87,11 +93,17 @@ func (c *client) ConnectHostnameToDeployment(ctx context.Context, directive ctyp
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        ingressName,
 			Labels:      labels,
-			Annotations: kubeNginxIngressAnnotations(directive),
+			Annotations: kubeNginxIngressAnnotations(directive, c.env),
 		},
 		Spec: netv1.IngressSpec{
 			IngressClassName: &ingressClassName,
 			Rules:            rules,
+			TLS: []netv1.IngressTLS{
+				{
+					Hosts:      []string{directive.Hostname},
+					SecretName: fmt.Sprintf("%s-tls", ingressName),
+				},
+			},
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


cluster/kube/client.go
```diff
@@ -752,3 +755,14 @@ func (c *client) deploymentsForLease(ctx context.Context, lid mtypes.LeaseID) (m
...
+func environmentVariablesToMap() map[string]string {
+	m := make(map[string]string)
+	for _, e := range os.Environ() {
+		if i := strings.Index(e, "="); i >= 0 {
+			m[e[:i]] = e[i+1:]
+		}
+	}
+
+	return m
+}
```
boz: Can we put this in a util directory somewhere?

cluster/kube/client_ingress.go
```diff
@@ -68,6 +68,12 @@ func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDir
 		}
 	}
 
+	if env["AKASH_PROVIDER_ISSUER_TYPE"] == "cluster-issuer" {
+		result["cert-manager.io/cluster-issuer"] = env["AKASH_PROVIDER_ISSUER_NAME"]
+	} else if env["AKASH_PROVIDER_ISSUER_TYPE"] == "issuer" {
+		result["cert-manager.io/issuer"] = env["AKASH_PROVIDER_ISSUER_NAME"]
+	}
+
```
boz: would prefer to have these as constants, similar to [this](https://github.com/akash-network/provider/blob/e7aa0b5b81957a130f1dc584f335c6f9e41db6b1/cluster/kube/builder/netpol.go#L39-L40)

boz (changes requested):
Looking great

cluster/util/environment.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import (
+	"os"
+	"strings"
+)
+
+func EnvironmentVariablesToMap() map[string]string {
+	m := make(map[string]string)
```
boz: `make(map[string]string,len(os.Environ()))` would be a bit better here (pre-allocated `map`).

boz (dismissed):
:rocket:

cluster/util/environment.go
cloud-j-luna: Totally, thank you!

boz (commented):
Just a note that it's generally preferable to use a framework for reading environment variables.   [This](https://github.com/caarlos0/env) or `cobra`+`viper` are good options.

boz (approved):
:rocket:

anilmurty:
from sig-support: @troian holding off merging this until he's confident about the code base from the changes for API refactor and 0.24 network upgrade

troian:
@cloud-j-luna rebase it on to `gpu` branch. we will merge it there first

cloud-j-luna:
@troian rebased in to `gpu` there are some checks failing regarding commit message but as the commits get squashed it shouldn't be a problem, right?

troian:
@cloud-j-luna the gpu branch will not be squashed. the check is failing as commit message needs to be formatted with conventional commits spec.

cloud-j-luna:
> @cloud-j-luna the gpu branch will not be squashed. the check is failing as commit message needs to be formatted with conventional commits spec. 

I meant this PR to be squashed

troian:
yeah, however, the commit message is taken from PR.
there are other tests failing too

cluster/mocks/client.go
troian, line 1: mocks need to be generated with `make codegen` command

cluster/kube/client_ingress.go
```diff
@@ -26,10 +26,11 @@ const (
 	akashIngressClassName = "akash-ingress-class"
 )
 
-func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective) map[string]string {
+func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective, env map[string]string) map[string]string {
 	// For kubernetes/ingress-nginx
 	// https://github.com/kubernetes/ingress-nginx
 	const root = "nginx.ingress.kubernetes.io"
+	const certManager = "cert-manager.io"
```
troian: move const up to the top section of the file

cluster/kube/client_ingress.go
```diff
@@ -66,11 +67,20 @@ func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDir
 		}
 	}
 
+	switch env["AKASH_PROVIDER_ISSUER_TYPE"] {
+	case "cluster-issuer":
+		result[fmt.Sprintf("%s/cluster-issuer", certManager)] = env["AKASH_PROVIDER_ISSUER_NAME"]
```
troian: 1. it looks like we don't need both `AKASH_PROVIDER_ISSUER_TYPE` and `AKASH_PROVIDER_ISSUER_NAME` and call it `AKASH_PROVIDER_CERT_ISSUER`
2. `AKASH_PROVIDER_CERT_ISSUER` should be defined in the provider env section
3. `env["AKASH_PROVIDER_ISSUER_NAME"]` may return `""` if env is not set

cluster/kube/client_ingress.go
```diff
@@ -66,11 +67,20 @@ func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDir
...
+		break
+	case "issuer":
+		result[fmt.Sprintf("%s/issuer", certManager)] = env["AKASH_PROVIDER_ISSUER_NAME"]
```
troian: comment from section above applies here as well

troian (changes requested):
I gave it a little thought and a few things must be addressed.
 - tlsEnabled should also account if deployment requested issuing TLS certificate, it does not have to be mandatory
 - all the changes need tests, including e2e

cluster/kube/client_ingress.go
cloud-j-luna: 1. We need both as the resulting label is different. This will also help in the future if we want to automate per-deployment/account issuers.

troian: what are the values of both env variables in this case?

cloud-j-luna: In this case it would be `AKASH_PROVIDER_ISSUER_TYPE` = "cluster-issuer", `AKASH_PROVIDER_ISSUER_NAME` would be whatever the provider chose as the cert-manager ClusterIssuer name on the cluster. By default the name is `letsencrypt` as the default issuer is Let's Encrypt.

troian: By having `letsencrypt` value, the code in the provider can deduct, that issue is `cluster-issuer` ?

cloud-j-luna: it can't thus the `cluster-issuer` in the annotation, so the provider knows that its a ClusterIssuer called letsencrypt rather than a namespace scoped issuer called letsencrypt as well

troian: if workload asks the provider to issue a certificate wouldn't it always be the `cluster-issuer`?

cluster/kube/client.go
```diff
@@ -98,6 +100,7 @@ func NewClient(ctx context.Context, log log.Logger, ns string, configPath string
 		ns:                ns,
 		log:               log.With("client", "kube"),
 		kubeContentConfig: config,
+		env:               util.EnvironmentVariablesToMap(),
```
troian, line 103: env variables must be validated for values and then read into dedicated struct fields

cluster/kube/client_ingress.go
```diff
@@ -26,10 +26,11 @@ const (
 	akashIngressClassName = "akash-ingress-class"
 )
 
-func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective) map[string]string {
+func kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective, env map[string]string) map[string]string {
```
troian:
```go
func (c *client) kubeNginxIngressAnnotations(directive ctypes.ConnectHostnameToDeploymentDirective) map[string]string
```
and use client struct instead of env variables

cluster/kube/client_ingress.go
cloud-j-luna: Depends on the provider configuration. For now only cluster-issuer are allowed for providers, but the `client_ingress.go` is ready to work with namespaced issuers.

cluster/kube/client.go
cloud-j-luna: Here is a direct conversion from the environment variables to a `map[string]string`. I can make validations but after this call. To have more fine-grained error handling on unexpected values.

troian: we would rather want to see dedicated struct fields instead of env converted to map. map is useful as intermediate store in some circumstances (like converting data to json). `Go` is a strict type of language and we use as much of it.

client instantiation is the last step where all configuration values must be validated

I'm not sure what `To have more fine-grained error handling on unexpected values.` means. If there is an unexpected configuration client should exit with error.

cloud-j-luna: It means we know the context in which the usage gave an error. Having a struct makes absolute sense! I'll change the code to support such configuration rather than environment variables

troian: We treat an error as an error, it is really hard to chase configuration errors on the fly.
1. the client must not interact with env variables, configs etc, the only way for it to receive configuration is thru a parameter to [NewClient](https://github.com/akash-network/provider/blob/718c95477f7be957474665a70391546cacf3185d/cluster/kube/client.go#L67).
2. Each environment variable must have a corresponding cli flag for it. That's why using env within the client breaks it.
3. Using environment variables inside implementation complicates writing unit tests.

remove environment variables from the client completely and use struct fields as values during annotations init.
I'm still unsure if using two different vars is the correct way to do it. if the workload going to use ns issuer then it has nothing to do with the cluster

troian:
apparently gh closes PRs if base branched merged, you’ll have to reopen it.

cloud-j-luna:
> apparently gh closes PRs if base branched merged, you’ll have to reopen it.

I have to open a new PR, to which branch do you want me to merge to? This branch is rebased on the `gpu` branch.

troian:
to the `main`

PR #112, opened 2023-04-18
feat: allow cert-manager annotations on ingress based on environment variables PT.2

cloud-j-luna:
This PR allows cert-manager annotations on ingress based on environment variables.


cloud-j-luna:
Previously closed issue #86

cluster/util/environment.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import (
+	"os"
+	"strings"
+)
+
+func EnvironmentVariablesToMap() map[string]string {
+	m := make(map[string]string, len(os.Environ()))
```
troian: hi,
as described in the previous PR, all configurations have to be done before the client is instantiated
and struct fields instead of map should be used

cloud-j-luna: Working on this right now. Writing a couple of tests and I'll push the changes.

cluster/kube/config.go
```diff
@@ -11,27 +12,41 @@ const (
...
 // configFromEnv creates a new clientConfig from environment variables.
 func configFromEnv() (*clientConfig, error) {
-	issuerType, ok := os.LookupEnv(akashProviderIssuerTypeStr)
-	if !ok || (issuerType != issuer && issuerType != clusterIssuer) {
-		return nil, errors.New(fmt.Sprintf("Invalid value for %s: %s", akashProviderIssuerTypeStr, issuerType))
-	}
+	sslEnabled := os.Getenv(akashProviderSslEnabledStr)
+	var sslCfg ssl
+
+	if sslEnabled != "" && sslEnabled != "0" {
```
troian: `bool`?

operator/hostnameoperator/hostname_operator.go
```diff
@@ -389,7 +391,7 @@ func (op *hostnameOperator) applyAddOrUpdateEvent(ctx context.Context, ev ctypes
 		if shouldConnect {
 			op.log.Debug("Updating ingress")
 			// Update or create the existing ingress
-			err = op.client.ConnectHostnameToDeployment(ctx, directive)
+			err = op.client.ConnectHostnameToDeployment(ctx, directive, op.env["AKASH_SSL_ENABLED"] != "")
```
troian: there is no need to define env variables manually. with such approach there is no possibility to use flags or config.

the configuration steps must be completely removed from the client.
the good starting point is [provider run](https://github.com/akash-network/provider/blob/main/cmd/provider-services/cmd/run.go)

troian:
i'll summarize requests here so it would be easier to follow
- we use viper as interface to parse and prevalidate configuration parameter that can be defined using:
  - default values
  - config file
  - env variables
  - flags
- don't define env variables, but flags instead, each flag will have env variable automatically by Viper (refer to it's documentation if you are eager to know how it works). for example `--home` flag will have `AKASH_HOME` env variable bound to it.
- the configuration must be evaluated and validated within command handler and all subsequent instantiation must use go types corresponding to its values. think it this way: type safety is a top priority
- if value can take only two states then `bool` would be the correct type. for example
instead
```go
// sslEnable is a string type
if sslEnable != "" && sslEnable == "1" {
```
do
```golang
// sslEnable is a bool type
if sslEnable {
```

[provider run](https://github.com/akash-network/provider/blob/main/cmd/provider-services/cmd/run.go) would be a good place to start

cluster/kube/config.go
cloud-j-luna: I got the value from `os.Getenv()` function. It comes as a string which can be empty or have value "0" for false.

cluster/util/environment.go
troian: ✂️

operator/hostnameoperator/hostname_operator.go
troian: `op.tlsEnabled` instead `op.env["AKASH_SSL_ENABLED"]`

troian:
i marked it as WIP until all issues are resolved. feel free to remove when done

_run/common-helm.mk
```diff
@@ -21,7 +21,7 @@ kind-install-helm-chart-loki:
 	helm upgrade --install promtail grafana/promtail \
 		--version $(PROMTAIL_VERSION) \
 		--namespace loki-stack \
-		-f ../promtail-values.yaml
+		-f ../promtail-values.yamlk
```
ghost, line 24: should it be `yamlk`?

PR #212, opened 2024-03-14
Lease service: `SendManifest`

### LINKED ISSUE #191: implement send manifest via GRPC endpoint
### Is your feature request related to a problem? Please describe.

Provider API is being migrated from REST to GRPC.
V1 version of lease API is defined in the [PR](https://github.com/akash-network/akash-api/pull/130)

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

n/a

### Search

- [X] I did search for other open and closed issues before opening this

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

_No response_

andrewhare:
Implements `SendManifest` from https://github.com/akash-network/akash-api/pull/130.

Resolves https://github.com/akash-network/support/issues/191

A few other things that are in this:

* A simple gRPC test harness. @troian mentioned that there are other gRPC tests, but I used this to familiarize myself with all the different interfaces behind the service, how they interact, and what they actually call.
* I refactored the mTLS logic into a single function. Eventually, the REST interface will be gone, but while we maintain both servers, we won't have multiple implementations to maintain as well.
* A small refactor to the `grpc.Serve` function to make it easier to test.
* I broke out the implementation of each proto service into its own file and its own struct. The main server struct composes these.


gateway/grpc/server.go
```diff
@@ -1,96 +1,38 @@
...
-	}), grpc.ChainUnaryInterceptor(mtlsInterceptor()))
-
-	pRPC := &grpcProviderV1{
-		ctx:    ctx,
-		client: client,
-	}
-
-	providerv1.RegisterProviderRPCServer(grpcSrv, pRPC)
-	gogoreflection.Register(grpcSrv)
+	var (
```
troian: lets not combine vars into section within function. it creates perception of grouped things that are not actually grouped.

gateway/grpc/server.go
```diff
@@ -114,102 +56,89 @@ func NewServer(ctx context.Context, endpoint string, certs []tls.Certificate, cl
 	return nil
 }
 
-func mtlsInterceptor() grpc.UnaryServerInterceptor {
-	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
+var (
```
troian: try keep global vars and types at the top of the file

gateway/grpc/server.go
```diff
@@ -114,102 +56,89 @@ func NewServer(ctx context.Context, endpoint string, certs []tls.Certificate, cl
...
+	_ providerv1.ProviderRPCServer = (*server)(nil)
+	_ leasev1.LeaseRPCServer       = (*server)(nil)
+)
+
+type server struct {
```
troian: define types at the top of the file

gateway/grpc/server.go
```diff
@@ -114,102 +56,89 @@ func NewServer(ctx context.Context, endpoint string, certs []tls.Certificate, cl
...
+				}
 
+				if owner != nil {
 					ctx = ContextWithOwner(ctx, owner)
 				}
 			}
 		}
 
-		return handler(ctx, req)
+		return h(ctx, req)
```
troian: `next` wound be a good example of var name here instead of `h`

gateway/utils/utils.go
```diff
@@ -22,68 +23,96 @@ func NewServerTLSConfig(ctx context.Context, certs []tls.Certificate, cquery cty
...
 	}
 
 	return cfg, nil
 }
+
+type certChain interface {
+	*x509.Certificate | []byte
+}
+
+func VerifyCertChain[T certChain](
```
troian: 1. do we really need generics here?
2. if Owner address is returned then better name would be `VerifyOwnerCert`. Overall issue with this approach is that some GRPC methods may not provider client cert. `Provider status` is good example. So assumption is that returned `sdk.Address` is never `nil`, which is not true in this case

gateway/rest/client.go
```diff
@@ -293,63 +296,18 @@ func (c *client) verifyPeerCertificate(certificates [][]byte, _ [][]*x509.Certif
...
-		Roots:                     certPool,
-		CurrentTime:               time.Now(),
-		KeyUsages:                 []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
-		MaxConstraintComparisions: 0,
+		return err
 	}
 
-	if _, err = cert.Verify(opts); err != nil {
-		return errors.Wrap(err, "tls: unable to verify certificate")
+	if !c.addr.Equals(prov) {
```
troian, line 309: shouldn't it be in verifyChain func?

gateway/grpc/server.go
```diff
@@ -114,102 +56,89 @@ func NewServer(ctx context.Context, endpoint string, certs []tls.Certificate, cl
...
-				return err
-			}
+// TODO(andrewhare): Possibly replace this with
+// https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors/logging
+// to get full request/response logging?
+func errorLogInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req any, i *grpc.UnaryServerInfo, h grpc.UnaryHandler) (any, error) {
+		resp, err := h(ctx, req)
+		if err != nil {
+			fromctx.LogcFromCtx(ctx).Error(i.FullMethod, "err", err)
```
troian: requesting object like log from context within interceptor is probably not good idea from performance prospective.
lets make `errorLogInterceptor` member of `server` and instantiate log during server setup.

`h` -> `next`

cmd/provider-services/cmd/run.go
```diff
@@ -746,7 +746,9 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	err = gwgrpc.NewServer(ctx, grpcaddr, []tls.Certificate{tlsCert}, service)
+	ctx = gwgrpc.ContextWithQueryClient(ctx, cl.Query())
+
+	err = gwgrpc.Serve(ctx, grpcaddr, []tls.Certificate{tlsCert}, service)
```
troian, line 752: I see were you going with it, however, this approach has a few caveats:
- there is error check prior serve starts.
- calling `Serve` makes assumptions everything is setup up, which is not true in this case, see above

I think better approach would be to `NewServer` as is, or

remove both `group.Go` from `NewServer` and put them into `Serve`. that implies some extra check as well as `Serve` to be blocking and error check by caller

troian (changes requested):

gateway/utils/utils.go
andrewhare: 1. It depends. It buys us 1) the ability to do the length checks before type assertion and 2) provide a sensible union type to the caller so they know what to pass. We can get rid of the generics once we remove the REST gateway (it's the only code passing in `[][]byte`. Another option would be to create wrapper functions that are typed to `[][]byte` and `[]*x509.Certificate` (but this would require duplicating the length checks). I'm not a huge fan of Go's generics, but it does make this particular code DRY.
2. That's a good point, I like that name better and thanks for the context.

andrewhare: Also, it occurs to me that the type constraint interface should be called `cert`, not `certChain`.

gateway/rest/client.go
andrewhare: Originally, I put it there, but the mTLS implementation for gRPC didn't have that check, so I moved it outside. What would the equivalent of `c.addr` be for the gRPC server if that was to check for equality as well?

cmd/provider-services/cmd/run.go
andrewhare: I'm not 100% sure what you mean. `Serve` currently has all the logic for running the server. The only change that affects `run.go` is that the function name changed. The implementation should be the same with a sub function (`newServer`) that returns the gRPC server to make testing easier.

In other words, I think I've already implemented it like you asked.

gateway/utils/utils.go
troian: lets remove generics from this usecase, `[][]byte` `[]*x509.Certificate` would expect that first expects valid encoded certificate. if there is need for `[][]byte` to exist wrapper would be clear and better option

gateway/rest/client.go
andrewhare: Would it just be whatever `net.Listener.Addr()` is set to for the server itself?

troian: those are akash address (aka account), not network

andrewhare: Right, good call 👍

PR #229, opened 2024-04-11
feat(private-containers): private registries

boz:
refs akash-network/support#54


troian:
merged akash-api secrets and [v0.0.62](https://github.com/akash-network/akash-api/releases/tag/v0.0.62) has it

PR #242, opened 2024-05-09
use reservation resource units if they exist

andrewhare:
Resolves https://github.com/akash-network/support/issues/166

This PR extends the `Request` struct to include resource units from a reservation (if it exists). The `CalculatedPrice` function will use those units if they are available.


go.mod
```diff
@@ -49,9 +49,7 @@ require (
 	sigs.k8s.io/kind v0.20.0
 )
 
-retract (
-	v0.6.0
-)
+retract v0.6.0
```
troian: no need to change this

bidengine/shellscript.go
```diff
@@ -126,42 +127,11 @@ func parseStorage(resource atypes.Volumes) []storageElement {
 	return res
 }
 
-func (ssp shellScriptPricing) CalculatePrice(ctx context.Context, req Request) (sdk.DecCoin, error) {
-	buf := &bytes.Buffer{}
+func (ssp shellScriptPricing) CalculatePrice(ctx context.Context, req Request, res ctypes.Reservation) (sdk.DecCoin, error) {
```
troian: there is no need to pass `ctypes.Reservation` as parameter here
instead, include allocated resourceUnits inside Request and use it instead of `for i, group := range req.GSpec.Resources`

bidengine/shellscript.go
```diff
@@ -126,42 +127,11 @@ func parseStorage(resource atypes.Volumes) []storageElement {
...
-			Count:            groupCount,
-			EndpointQuantity: endpointQuantity,
-			IPLeaseQuantity:  util.GetEndpointQuantityOfResourceUnits(group.Resources, atypes.Endpoint_LEASED_IP),
-		}
-	}
-
-	encoder := json.NewEncoder(buf)
-	err := encoder.Encode(dataForScript)
-	if err != nil {
+	var buf bytes.Buffer
```
troian: leave it as `buf := &bytes.Buffer{}`

troian (changes requested):

go.mod
andrewhare: This happened automatically with `go mod tidy`. I can roll it back.

bidengine/pricing.go
```diff
@@ -317,9 +318,11 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.DecCoin, sdk.DecCoin) {
...
 }
 
-var errPathEmpty = errors.New("script path cannot be the empty string")
-var errProcessLimitZero = errors.New("process limit must be greater than zero")
-var errProcessRuntimeLimitZero = errors.New("process runtime limit must be greater than zero")
+var (
+	errPathEmpty               = errors.New("script path cannot be the empty string")
+	errProcessLimitZero        = errors.New("process limit must be greater than zero")
+	errProcessRuntimeLimitZero = errors.New("process runtime limit must be greater than zero")
+)
```
andrewhare, line 325: Most of these changes are happening automatically [because I'm running `gofumpt`](https://go.googlesource.com/tools/+/refs/heads/master/gopls/doc/settings.md#gofumpt-bool). Let me know if this makes the diffs too noisy.

bidengine/shellscript_test.go
troian, line 1: `pricing_test.go` has tests for both scale and script, lets keep script pricing there

troian (approved):

PR #243, opened 2024-05-21
fix(gateway/rest): use cluster status instead of manifest

troian:
check service status prior trying shell using cluster API refs akash-network/support#87


troian:
superseded by #247

PR #263, opened 2025-01-13
chore: fix function name and struct field name in comment

dashangcun:


troian:
this is codegenerated code

PR #273, opened 2025-03-05
Add dynamic ingress hostname and NodePort env vars - v2

pcfreak30:
This version is a complete redesign in approach and requires sourcing in bash a `config.env` file which is added to a `emptydir` mount.

The configurations are discovered via an init container added to every deployment, be it a `statefulset` or `deployment` type.

The default `serviceaccount` is modified to add a RBAC rule to ensure the init container can read the data needed to write the file.

Several experiments were tested to try and make everything be auto injected as ENV's but it was too brittle and required knowledge of the images themselves which the provider does **not** have at the time of creation of k8 resources.


cloud-j-luna:
Could you please rebase this branch?

cluster/kube/builder/config.go
```diff
@@ -0,0 +1,125 @@
...
+	AkashConfigMount    = "/akash-cfg"
+	AkashConfigInitName = "akash-init"
+	AkashConfigEnvFile  = "config.env"
+
+	// RBAC constants
+	AkashRoleName    = "akash-role"
+	AkashRoleBinding = "akash-binding"
+
+	// Init container script
+	akashInitScript = `
```
cloud-j-luna, line 15: This adds a lot of complexity to the code. I believe you could achieve this using the standard library and the Kubernetes API which would result in a more maintainable PR

pcfreak30: @cloud-j-luna 

From what I found, the only way to access the data is AFTER the container is created which my 1st attempt meant calling creation twice since the API decides on the nodeports... so you end up needing an INIT script to query that data.

I also looked at the kube meta yaml stuff and theres no way to inject the node port info in...

pcfreak30:
> Could you please rebase this branch?

done.

cluster/kube/builder/config.go
cloud-j-luna: Why is accessing the data after the container is created through kube api not an option? I'm not a fan of the idea of injecting an init container into deployments that run an embedded shell script from the provider... worst case scenario would at least mount the script instead.

pcfreak30: I do not know the possibility if you can access the meta api from within the main container.  My original PR was a lot more buggy and involved doing 2 creates as I stated because you had to get the info from kube since kube decides the node ports.

The big question I guess is when/how can a container get information about the nodeports the kube daemon assigns to it. This can def be tricky since it goes to a service entity and not the container itself, IIRC?

I found the "kube yaml meta" cant do it, the kube api decides it, and you have to add in RBAC to enable api access. I don't know if you can do so in the main container like how EC2 can access AWS metaservice API's.

PR #280, opened 2025-03-19
Update node-discovery.go

deathlessdd:
Was going though some of the issues and I remember that 63char was to long. I dont own a super card but was wondering if this code will work. We could technically add all of the gpus with long names to be able to rename them. In theory this is simple enough. Hopefully this can help.


deathlessdd (commented):
updated

troian:
hi @deathlessdd, thank you for bringing this up.
 we're going to implement the use of hashes instead of labels, which will give predictable 32-character unique names.

unfortunately this is rather hack and we've cannot incorporate it in to the code

PR #310, opened 2025-07-30
fix: 295: Shell access to a SHM mounted workload

### LINKED ISSUE #295: Shell is not accessible when shm is enabled.
I follwed https://akash.network/docs/getting-started/stack-definition-language/#shared-memory-shm and https://gist.github.com/chainzero/0dea9f2e1c4241d2e4d490b37153ec86 page. I've tried multiple providers, but they didn't work.

**Describe the bug**
A clear and concise description of what the bug is. Please include the following:
- account address = akash1tn8pwdv2c5qlxp3m6pcq7q4k4py2ca40zl46es
- deployment sequence =
```
---
version: "2.0"
services:
  service-1:
    image: ubuntu:22.04
    expose:
      - port: 80
        as: 80
        to:
          - global: true
    params:
      storage:
        shm:
          mount: /dev/shm
    command:
      - bash
      - "-c"
    args:
      - >-
        apt-get update;apt-get install nano; sleep infinity
profiles:
  compute:
    service-1:
      resources:
        cpu:
          units: 12
        memory:
          size: 32Gb
        storage:
          - size: 40Gi
          - name: shm
            size: 2Gi
            attributes:
              persistent: false
              class: ram
        gpu:
          units: 1
          attributes:
            vendor:
              nvidia:
                - model: p100
                - model: v100
                - model: rtxa6000
                - model: rtx3070
                - model: rtx3060
                - model: rtx3090
                - model: rtx8000
                - model: rtx4090
                - model: t4
                - model: a6000
  placement:
    dcloud:
      pricing:
        service-1:
          denom: uakt
          amount: 10000

deployment:
  service-1:
    dcloud:
      profile: service-1
      count: 1
```

- deploy.yaml
- group sequence if there is more than one group

**To Reproduce**
1. deploy  sdl yaml (enabled shm)
2.  click shell

**Expected behavior**
when shm is enabled, shell doesn't work.

**Screenshots**

Image

Image

zach-source:
Closes https://github.com/akash-network/support/issues/295

This commit addresses an issue with the provider when attempting to SSH to a lease with a shared memory mount. It also corrects a websocket protocol violation on error (encountered while debugging the SHM issue). The core issue essentially was when the lease-shell command attempted to connect, it would try to do so on a Statefulset due to detecting a volume on the deployment. Having it follow the convention of the Deploy() function let's it connect to the deployment properly.

1. StatefulSet Detection Fix:
   - Fixed incorrect logic in ServiceStatus that was checking for any mounted storage instead of persistent storage to determine workload type
   - The bug caused services with RAM volumes to incorrectly attempt StatefulSet operations, resulting in "statefulsets.apps not found" errors
   - Now correctly checks storage.Attributes.Find(sdl.StorageAttributePersistent) to match the deployment creation logic

2. WebSocket Error Handling in lease-shell:
   - Fixed improper error handling when ServiceStatus fails during shell operations
   - Previously used http.Error() on WebSocket connections, causing protocol violations
   - Now properly uses WebSocket writer with LeaseShellCodeFailure and logs errors
   - Prevents silent failures and improves debugging for shell access issues

Added comprehensive unit tests for StatefulSet detection logic covering:
- Services with persistent storage (should use StatefulSet)
- Services with non-persistent storage (should use Deployment)
- Services with no storage (should use Deployment)

These fixes ensure proper workload type detection and improve error visibility for lease-shell operations, resolving issues with shell access to deployments using RAM volumes.

### DIFF
--- a/cluster/kube/client.go
+++ b/cluster/kube/client.go
@@ -935,13 +935,15 @@ func (c *client) ServiceStatus(ctx context.Context, lid mtypes.LeaseID, name str
 		return nil, kubeclienterrors.ErrNoServiceForLease
 	}
 
+	// Check if this service uses persistent storage (should be StatefulSet)
+	// This logic must match the deployment creation logic in Deploy()
 	isDeployment := true
-	if params := svc.Params; params != nil {
-		for _, param := range params.Storage {
-			if param.Mount != "" {
-				isDeployment = false
-				break
-			}
+	persistent := false
+	for i := range svc.Resources.Storage {
+		attrVal := svc.Resources.Storage[i].Attributes.Find(sdl.StorageAttributePersistent)
+		if persistent, _ = attrVal.AsBool(); persistent {
+			isDeployment = false
+			break
 		}
 	}
 
--- a/cluster/kube/client_test.go
+++ b/cluster/kube/client_test.go
@@ -7,6 +7,7 @@ import (
 	manifest "github.com/akash-network/akash-api/go/manifest/v2beta2"
 	mtypes "github.com/akash-network/akash-api/go/node/market/v1beta4"
 	types "github.com/akash-network/akash-api/go/node/types/v1beta3"
+	"github.com/akash-network/node/sdl"
 	"github.com/akash-network/node/testutil"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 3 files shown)


Zblocker64:
closing in favor of https://github.com/akash-network/provider/pull/421

PR #318, opened 2025-09-02
feat: external ports reservation

vertex451:
# Port Reservation System for Akash Provider

Resolves https://github.com/akash-network/support/issues/234

Important limitation: we return only a list of ports because when an order is created, the provider sees only limited data about future deployment, not the full manifest. 

## Testing
**Manual E2E Testing:**
I spin up Aakash blockchain and run a provider with my changes.
Then I created a deployment, which triggers bidding, and port reservation happens.
So, when the bid from the provider is here, we can query `provider-bid-proposed-ports`:
```
{"proposed_ports":[30106,30107]}
```
Now we can create a lease, send a manifest, and query provider lease status for that deployment, and in `forwarded_ports` we can see that the service uses ports reserved during the bidding phase:
```
  "forwarded_ports": {
    "bew": [
      {
        "host": "localhost",
        "port": 80,
        "externalPort": 30106,
        "proto": "TCP",
        "name": "bew"
      }
    ],
    "web": [
      {
        "host": "localhost",
        "port": 80,
        "externalPort": 30107,
        "proto": "TCP",
        "name": "web"
      }
    ]
  },
  ```

## Overview
Enables users to query proposed external ports from providers during the bidding phase, allowing informed provider selection based on port allocations.

## Problem 

Users only saw ports after lease creation, making provider comparison impossible.

## Solution 
1. Reserve ports during bidding.
2. Using those ports if a lease is created
3. Releasing ports on corresponding events
4. HTTP endpoint to query proposed allocations.

## Architecture
- **HTTP API**: `/bid-proposed-ports/{owner}/{dseq}/{gseq}/{oseq}`. Transport layer.
- **`PortManager`**: Unified port operations (order + lease phases). Usecase layer.
- **`portreserve.Store`**: separated db layer

### Port Release Triggers
1. **Bid Lost**: Order ports released immediately when the bid doesn't win
2. **TTL Expiration**: Auto-cleanup after 5 minutes (orders) or 1 minute (leases)
3. **Lease Closed**: All lease ports are released when the lease terminates
4. **Manual Cleanup**: Periodic cleanup routine removes expired reservations
5. **Provider Restart**: Stale reservations cleaned on startup recovery

## TODOs
- [ ] persistent store  
- [ ] Configurable port ranges and TTL
- [ ] Better testing with concurrent deployments

## Proposal: Persistent Port Reservations via CRD

Memory-only port reservations are lost on provider restart. Introduce a lightweight PortReservation CRD:

Fields: port, bidID, leaseID, status (reserved/allocated/released), expiresAt.

On bid → create reserved; bid lost/TTL → released; bid won → allocated.

Manifest reads CRDs for port assignment.

Survives provider restarts, no external DB needed.

### DIFF
--- a/cluster/inventory.go
+++ b/cluster/inventory.go
@@ -138,8 +138,29 @@ func newInventoryService(
 		waiter:                 waiter,
 	}
 
-	// Create centralized port manager with shared store
-	is.portManager = NewPortManager(is.log, portreserve.GetSharedStore())
+	// Create centralized port manager with shared store and validated port range
+	store := portreserve.GetSharedStore()
+
+	// Validate port range configuration
+	minPort := int32(30100) // Default minimum
+	maxPort := int32(32767) // Default maximum
+
+	// TODO: Read from config when port range configuration is added
+	// For now, use safe defaults
+
+	if minPort < 1 || minPort > 65535 {
+		return nil, fmt.Errorf("invalid minimum port %d: must be between 1-65535", minPort)
+	}
+	if maxPort < 1 || maxPort > 65535 {
+		return nil, fmt.Errorf("invalid maximum port %d: must be between 1-65535", maxPort)
+	}
+	if minPort > maxPort {
+		return nil, fmt.Errorf("invalid port range: minimum port %d > maximum port %d", minPort, maxPort)
+	}
+
+	// Set validated port range on store
+	store.SetPortRange(minPort, maxPort)
+	is.portManager = NewPortManager(is.log, store)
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: _run/kube/deployment.yaml [hunk 1]
```diff
@@ -6,22 +6,14 @@ services:
     image: quay.io/ovrclk/demo-app
     expose:
       - port: 80
-        as: 80
```
vertex451, line 9: I kept this diff to show example of deployment i used to create NodePort for randpom port assigments

PR #319, opened 2025-09-04
feat(deployment): enable AutomountServiceAccountToken for the log-collector image

ygrishajev:


cluster/kube/builder/deployment.go
```diff
@@ -79,16 +84,51 @@ func (b *deployment) Create() (*appsv1.Deployment, error) { // nolint:golint,unp
...
+	image := container.Image
+	automountServiceAccountToken := b.determineAutomountServiceAccountToken(image)
+	uobj.Spec.Template.Spec.AutomountServiceAccountToken = &automountServiceAccountToken
+
 	return uobj, nil
 }
+
+func (b *deployment) determineAutomountServiceAccountToken(image string) bool {
+	automountImages := []string{
+		"ghcr.io/akash-network/log-collector",
```
cloud-j-luna, line 108: This should be configurable as any change to the image location would require recompilation.

cluster/kube/builder/deployment.go
```diff
@@ -79,16 +84,51 @@ func (b *deployment) Create() (*appsv1.Deployment, error) { // nolint:golint,unp
...
+	}
+
+	imageName := extractImageName(image)
+
+	for _, automountImage := range automountImages {
+		if strings.EqualFold(imageName, automountImage) {
```
cloud-j-luna, line 114: Should we have tags into consideration? The image set is using the latest tag but couldn't the log-collector be specified with a specific version?

ygrishajev: multiple cases are taken into account. added more test cases to demo https://github.com/akash-network/provider/pull/319/commits/9a25a8242af8185969a389a7af33691d5e275f5b

PR #334, opened 2025-10-13
fix: node env variable not being inherited on viper

cloud-j-luna:
Both `AP_NODE` and `AKASH_NODE` environment variables were not being parsed through viper to be used with the `provider-services` CLI. This prevented using `AKASH_NODE` to configure the node when creating a deployment.

This PR configures the CLI to accept both `AP_` and `AKASH_`.


troian (changes requested):
the binding is imported from the [node repo](https://github.com/akash-network/node/blob/4ab1cd969fd3bf20b9ecfc6d5889f2a9bd9ec5d1/util/cli/configs.go#L61)

PR #371, opened 2026-03-10
fix: bid is not closed after the manifest timeout

vertex451:
# Resolves:
https://github.com/akash-network/support/issues/435

# Root of the issue:
1. The broadcast of bid close runs with wd.ctx, but we call ShutdownInitiated before it finishes. 
2. That closes stoppingch, so WatchChannel returns and cancels wd.ctx while the broadcast is still in progress. 
3. The broadcast then fails with "context canceled" and the bid is not closed on-chain.

# Changes

1. Commit to the broadcast once started. Even if stop() signal received during the wait of the broadcast result - we don't cancel the broadcast.

I moved wd.lc.ShutdownInitiated(err) after `case result := <-runch:`, so it will be called right after runner will finish its job.

2. Added broadcast timeout, that will cancel the execution in FlagTxBroadcastTimeout and return.

3. I added a select case for the `case err = <-wd.lc.ShutdownRequest():` in case shutdown will be initiated outside to prevent deadlock if stop() is called during the broadcast.

# Is that a regression that we got recently?

The bug is long-standing. The v1 upgrade likely made it more visible by changing client behavior and timing, not by introducing new logic. So may this issue may appeared after pkg.akt.dev client changes. 

# Testing
Created a lease and waited 5min.
Provider logs:
```
5:32PM INF watchdog closing bid leaseID=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1/akash1ff4s67p9xjykezkn7gnmdzpz8vdw3kn0ewx7ss module=provider-manifest
5:32PM INF watchdog done lease=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7 module=provider-manifest
5:32PM INF lease closed lease=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1/akash1ff4s67p9xjykezkn7gnmdzpz8vdw3kn0ewx7ss module=provider-manifest
5:32PM INF unreserving unmanaged order cmp=service lease=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1/akash1ff4s67p9xjykezkn7gnmdzpz8vdw3kn0ewx7ss module=provider-cluster
5:32PM INF lease removed deployment=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7 lease=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1/akash1ff4s67p9xjykezkn7gnmdzpz8vdw3kn0ewx7ss module=manifest-manager
5:32PM DBG unreserving capacity cmp=inventory-service module=provider-cluster order=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1
5:32PM INF attempting to removing reservation cmp=inventory-service module=provider-cluster order=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1
5:32PM INF removing reservation cmp=inventory-service module=provider-cluster order=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1
5:32PM INF unreserve capacity complete cmp=inventory-service module=provider-cluster order=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/7/1/1
```

Also tested a cancelation timeout of 10ms:
```
5:24PM ERR failed closing bid err="context deadline exceeded" leaseID=akash1nfyvsjhl7wcyduq7hkx87jja7mu5qyfmcrf0sv/6/1/1/akash1ff4s67p9xjykezkn7gnmdzpz8vdw3kn0ewx7ss module=provider-manifest
```
Which is expected.

### DIFF
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -263,7 +263,7 @@ loop:
 				if ev.LeaseID.GetProvider() != s.session.Provider().Address().String() {
 					continue
 				}
-				s.session.Log().Info("lease won", "lease", ev.LeaseID)
+				s.session.Log().Info("handling lease", "lease", ev.LeaseID)
 				s.handleLease(ev, true)
 			case *dtypes.EventDeploymentUpdated:
 				s.session.Log().Info("update received", "deployment", ev.ID, "version", ev.Hash)
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -263,7 +263,7 @@ loop:
 				if ev.LeaseID.GetProvider() != s.session.Provider().Address().String() {
 					continue
 				}
-				s.session.Log().Info("handling lease", "lease", ev.LeaseID)
+				s.session.Log().Info("lease won", "lease", ev.LeaseID)
 				s.handleLease(ev, true)
 			case *dtypes.EventDeploymentUpdated:
 				s.session.Log().Info("update received", "deployment", ev.ID, "version", ev.Hash)
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -351,7 +351,7 @@ func (s *service) handleLease(ev event.LeaseWon, isNew bool) {
 	if isNew && s.config.ManifestTimeout > time.Duration(0) {
 		// Create watchdog if it does not exist AND a manifest has not been received yet
 		if watchdog := s.watchdogs[ev.LeaseID.DeploymentID()]; watchdog == nil {
-			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, s.config.ManifestTimeout)
+			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, s.config.ManifestTimeout, s.config.BroadcastTimeout)
 			s.watchdogs[ev.LeaseID.DeploymentID()] = watchdog
 		}
 	}
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -351,7 +351,8 @@ func (s *service) handleLease(ev event.LeaseWon, isNew bool) {
 	if isNew && s.config.ManifestTimeout > time.Duration(0) {
 		// Create watchdog if it does not exist AND a manifest has not been received yet
 		if watchdog := s.watchdogs[ev.LeaseID.DeploymentID()]; watchdog == nil {
-			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, s.config.ManifestTimeout, s.config.BroadcastTimeout)
+			timeout := manifestTimeoutForLease(s.config.ManifestTimeout, ev.CreatedAt, s.session.CreatedAtBlockHeight())
+			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, timeout, s.config.BroadcastTimeout)
 			s.watchdogs[ev.LeaseID.DeploymentID()] = watchdog
 		}
 	}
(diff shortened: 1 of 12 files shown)


gateway/rest/router.go
```diff
@@ -110,7 +110,7 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client, ctxCon
 		authorizeProviderMiddleware,
 		requireOwner,
 	)
-
```
vertex451: linter fails because of that locally

manifest/service.go
```diff
@@ -263,7 +263,7 @@ loop:
 				if ev.LeaseID.GetProvider() != s.session.Provider().Address().String() {
 					continue
 				}
-				s.session.Log().Info("lease won", "lease", ev.LeaseID)
+				s.session.Log().Info("handling lease", "lease", ev.LeaseID)
```
vertex451: we have duplicate of that message two times, I made the second one more presice.

cloud-j-luna (dismissed):
LGTM

manifest/service.go
troian: message comes from different modules, keep original

manifest/watchdog.go
```diff
@@ -75,17 +75,20 @@ func (wd *watchdog) run() {
...
 	if runch != nil {
-		result := <-runch
-		if err := result.Error(); err != nil {
-			wd.log.Error("failed closing bid", "err", err)
+		select {
+		case result := <-runch:
+			if err := result.Error(); err != nil {
+				wd.log.Error("failed closing bid", "err", err)
+			}
+		case err = <-wd.lc.ShutdownRequest():
```
troian, line 104: if err is not used in case then `_ = `

troian (approved):

troian (changes requested):
address request changes

manifest/watchdog.go
vertex451: The error is used in the very end of this func at line 93: `wd.lc.ShutdownInitiated(err)`
And it is visible outside `if runch != nil {` block because it is declared at function scope at line 65.

manifest/watchdog.go
```diff
@@ -75,17 +75,20 @@ func (wd *watchdog) run() {
...
+			// we need this case to skip waiting on runch when stop() is called
 		}
 	}
+	wd.lc.ShutdownInitiated(err)
```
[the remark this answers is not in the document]
vertex451: I did the change, and also I implemented the following rule - commit to the broadcast once started.
please check it out.

manifest/service.go
```diff
@@ -351,7 +351,7 @@ func (s *service) handleLease(ev event.LeaseWon, isNew bool) {
 	if isNew && s.config.ManifestTimeout > time.Duration(0) {
 		// Create watchdog if it does not exist AND a manifest has not been received yet
 		if watchdog := s.watchdogs[ev.LeaseID.DeploymentID()]; watchdog == nil {
-			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, s.config.ManifestTimeout)
+			watchdog = newWatchdog(s.session, s.lc.ShuttingDown(), s.watchdogch, ev.LeaseID, s.config.ManifestTimeout, s.config.BroadcastTimeout)
```
troian, line 354: if provider service restarts, does it start counting from 0, or takes into account when lease was created?

vertex451: No, the lease creation time is not taking into an account.
I can improve the solution by taking Lease.CreatedAt and computing the remaining time by converting currentBlock to the time(the same we did in isStaleBid)

Not doing this since we discussed the refactor of the serial broadcaster.

chalabi2: @troian 
The manifest timeout now respects provider restarts. Recovered leases carry their chain `CreatedAt` height through `LeaseWon`, and the watchdog computes the remaining timeout from the provider session height instead of starting a fresh full timeout window after restart.

Live lease-created events still get the normal full timeout because the provider sees them as they happen. Recovered active leases use the remaining time, and if the timeout already elapsed, the watchdog closes the bid immediately.

PR #373, opened 2026-03-18
fix(bidengine): close bid on EventGroupClosed when deployment closed

### LINKED ISSUE #438: Provider bid stays open when deployment is closed without lease
## Describe the bug

When a deployer closes a deployment on chain **without creating a lease**, the provider’s bid stays open. The provider should react to the closure and close the bid, but it does not.

**Root cause:** The chain emits `EventDeploymentClosed` and `EventGroupClosed` when a deployment is closed without a lease, but **not** `EventOrderClosed`. The provider only handled `EventOrderClosed`, so it never closed the bid in this case.

**Fix:** Handle `EventGroupClosed` in the bidengine order handler. When the event’s group matches the order’s group, the order is closed and the bid is cleaned up (unreserve, close bid).

## To Reproduce

1. Create a deployment (e.g. `akash tx deployment create deploy.yaml`).
2. Wait for providers to bid.
3. Close the deployment **before** creating a lease (`akash tx deployment close --dseq <dseq> --from <wallet> -y`).
4. Observe that the provider’s bid remains open instead of being closed.

## Expected behavior

When a deployment is closed without a lease, the provider should:
- Detect the closure via `EventGroupClosed`
- Close the order and bid
- Unreserve capacity

## Additional context

- **Provider component:** bidengine (`bidengine/order.go`)
- **Fix:** Add handling for `*dv1.EventGroupClosed` in the order event loop. When `ev.ID.Equals(o.orderID.GroupID())`, close the order and run the normal cleanup (unreserve, close bid).
- **Event flow:** On deployment close without lease, the chain emits `EventDeploymentClosed`, `EventGroupClosed`, and `EventBidClosed`. The provider must handle `EventGroupClosed` (or equivalent) to close bids in this scenario.

Zblocker64:
fixes https://github.com/akash-network/support/issues/438


cloud-j-luna (approved):

vertex451 (approved):

bidengine/order.go
```diff
@@ -294,14 +295,20 @@ loop:
...
 			case *mtypes.EventOrderClosed:
-				// different deployment
 				if !ev.ID.Equals(o.orderID) {
 					break
 				}
-
 				o.log.Info("order closed")
 				orderCompleteCounter.WithLabelValues("order-closed").Inc()
 				break loop
+			case *dv1.EventGroupClosed:
```
troian, line 308: deployment close triggers `EventOrderClosed`

Zblocker64: The chain emits `EventDeploymentClosed` and `EventGroupClosed` when a deployment is closed without a lease, but not `EventOrderClosed`. The provider only handled `EventOrderClosed`, so it never closed the bid in this case.

Should we look to fix this in node?

troian: if deployment is closed then order is closed as well.
create test deployment without lease, listen for events and close deployment. open issue if `EventOrderClosed` does not show up

Zblocker64: https://github.com/akash-network/support/issues/444

PR #379, opened 2026-03-23
Artem/fix gpu count

vertex451:
# Resolves: 
https://github.com/akash-network/support/issues/429

# Details 
I haven't found a bug in how do we handle inventory events.

Also, I've found recent report about GPU driver reporting wrong values.
- https://github.com/NVIDIA/k8s-device-plugin/issues/1645

That is why I decided to add a guardrails that prevents uint underflow, also I added logging, so we can track this in the future.

# Changes

- Safe int64→uint64 conversion - clamp negatives to 0 instead of raw cast; guard against -1 sentinel in Available() returning MaxInt64
- Sanitize all resource quantities at ingest - reject negative values from knode.Status.Allocatable/Capacity for CPU, Memory, GPU, Storage
- GPU allocatable upper-bound - clamp to Capacity if Allocatable > Capacity
- Pod key uses namespace/name - prevents cross-namespace accounting collision
- Handle watch.Added on node watcher reconnect - was silently dropped, leaving stale allocatable
- podsWatch.Stop() on restart - old code drained one event and leaked the watcher
- Precision-preserving Sub() - old Value() - Value() truncated milli-precision, drifting CPU allocated toward 0

# Testing
Tested locally at non-gpu machine and checked if CPU is consistently changed compared to the main branch.
Checked CPU and memory changes at each stages:
 Create a provider -> Create a deployment -> Create a lease -> Send a manifest -> Delete a deployment.

### DIFF
--- a/cluster/kube/operators/clients/inventory/client_test.go
+++ b/cluster/kube/operators/clients/inventory/client_test.go
@@ -317,16 +317,11 @@ func TestInventoryZero(t *testing.T) {
 	require.Len(t, inv.Metrics().Nodes, 0)
 }
 
-func TestInventoryMetrics_bad_gpu_values_no_overflow(t *testing.T) {
-	scaffold := makeInventoryScaffold(t)
-	cl, err := NewClient(scaffold.ctx)
-	require.NoError(t, err)
-	require.NotNil(t, cl)
-
-	scaffold.gInv.invch <- inventoryV1.Cluster{
+func makeBaseCluster() inventoryV1.Cluster {
+	return inventoryV1.Cluster{
 		Nodes: inventoryV1.Nodes{
 			inventoryV1.Node{
-				Name: "bad-gpu-node",
+				Name: "node",
 				Resources: inventoryV1.NodeResources{
 					CPU: inventoryV1.CPU{
 						Quantity: inventoryV1.NewResourcePairMilli(1000, 1000, 0, resource.DecimalSI),
@@ -335,7 +330,7 @@ func TestInventoryMetrics_bad_gpu_values_no_overflow(t *testing.T) {
 						Quantity: inventoryV1.NewResourcePair(1024, 1024, 0, resource.DecimalSI),
 					},
 					GPU: inventoryV1.GPU{
-						Quantity: inventoryV1.NewResourcePair(0, -1, 0, resource.DecimalSI),
+						Quantity: inventoryV1.NewResourcePair(0, 0, 0, resource.DecimalSI),
 					},
 					EphemeralStorage: inventoryV1.NewResourcePair(100, 100, 0, resource.DecimalSI),
 					VolumesAttached:  inventoryV1.NewResourcePair(0, 0, 0, resource.DecimalSI),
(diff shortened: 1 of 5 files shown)


operator/inventory/node-discovery.go
```diff
@@ -733,29 +761,30 @@ func addPodAllocatedResources(node *v1.Node, pod *corev1.Pod) {
 	}
 }
 
-func subAllocatedNLZ(allocated *resource.Quantity, val resource.Quantity) {
-	newVal := allocated.Value() - val.Value()
```
vertex451, line 737: .Value() truncates to whole units. For CPU (stored in milli), allocated = 500m means .Value() == 0, val = 200m means .Value() == 0, so newVal = 0 and you'd Set(0) - losing 300m of tracked allocation.

Over time, every pod delete that involved sub-unit CPU quantities would drift Allocated toward 0, making Available appear larger than reality.

On the other hand, .Sub() operates on the full-precision quantity

operator/inventory/node-discovery.go
```diff
@@ -419,10 +437,8 @@ func (dp *nodeDiscovery) monitor() error {
 
 	restartPodsWatcher := func() error {
 		if podsWatch != nil {
-			select {
-			case <-podsWatch.ResultChan():
```
vertex451, line 423: case <-podsWatch.ResultChan() only drained one buffered event and didn't stop the watcher

operator/inventory/node-discovery.go
```diff
@@ -341,6 +355,10 @@ initloop:
 	}
 }
 
+func podKey(pod *corev1.Pod) string {
+	return pod.Namespace + "/" + pod.Name
```
vertex451: We need this to prevent a collision between pods with the same name but from different namespaces.

operator/inventory/node-discovery.go
```diff
@@ -518,21 +534,22 @@ func (dp *nodeDiscovery) monitor() error {
...
-						if nodeAllocatableChanged(knode, obj) {
-							updateNodeInfo(obj, &node)
-							if err = restartPodsWatcher(); err != nil {
-								return err
-							}
+				if obj.Name != dp.name {
+					continue
+				}
+				switch evt.Type {
+				case watch.Added:
```
vertex451: This event was ignored before

cluster/kube/operators/clients/inventory/client_test.go
```diff
@@ -315,6 +317,128 @@ func TestInventoryZero(t *testing.T) {
...
+				c.Storage = inventoryV1.ClusterStorage{
+					{
+						Quantity: inventoryV1.NewResourcePair(0, -1, 0, resource.DecimalSI),
+						Info:     inventoryV1.StorageInfo{Class: "default"},
+					},
+				}
+			},
+			check: func(t *testing.T, m inventoryV1.Metrics) {
+				assert.Equal(t, uint64(0), m.TotalAllocatable.Storage["default"])
+			},
```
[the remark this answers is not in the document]
vertex451: Done

operator/inventory/node-discovery.go
cloud-j-luna: Perhaps we could use something along these lines, instead of implying the format ourselves.
```go
types.NamespacedName{
    Namespace: pod.Namespace,
    Name:      pod.Name,
}.String()
```

operator/inventory/node-discovery.go
```diff
@@ -518,21 +534,22 @@ func (dp *nodeDiscovery) monitor() error {
...
+					fallthrough
+				case watch.Modified:
```
cloud-j-luna, line 543: This seems unnecessary, consider:

```go
case watch.Added, watch.Modified:
```

operator/inventory/node-discovery.go
```diff
@@ -659,41 +677,62 @@ func (dp *nodeDiscovery) initNodeInfo(gpusIDs RegistryGPUVendors, knode *corev1.
...
+	// Reset all fields before repopulating from knode so that resources
+	// removed between updates (e.g. GPU device plugin unregistered) don't leave stale values.
+	node.Resources.CPU.Quantity.Allocatable.SetMilli(0)
+	node.Resources.CPU.Quantity.Capacity.SetMilli(0)
+	node.Resources.Memory.Quantity.Allocatable.Set(0)
+	node.Resources.Memory.Quantity.Capacity.Set(0)
+	node.Resources.EphemeralStorage.Allocatable.Set(0)
+	node.Resources.EphemeralStorage.Capacity.Set(0)
+	node.Resources.GPU.Quantity.Allocatable.Set(0)
+	node.Resources.GPU.Quantity.Capacity.Set(0)
```
cloud-j-luna, line 697: Can't you just set `Resources` to its zero-value?

operator/inventory/node-discovery.go
```diff
@@ -733,29 +772,30 @@ func addPodAllocatedResources(node *v1.Node, pod *corev1.Pod) {
...
-func subAllocatedNLZ(allocated *resource.Quantity, val resource.Quantity) {
-	newVal := allocated.Value() - val.Value()
-	if newVal < 0 {
-		newVal = 0
+func subAllocatedNLZ(log logr.Logger, nodeName, resourceName string, allocated *resource.Quantity, val resource.Quantity) {
+	before := allocated.DeepCopy()
+	allocated.Sub(val)
+	if allocated.Cmp(*quantityZero) < 0 {
+		log.Error(errAllocatedUnderflowClamp, "allocated underflow clamped", "node", nodeName, "resource", resourceName, "allocated_before", before.String(), "subtracted", val.String())
+		allocated.Set(0)
```
cloud-j-luna, line 780: This seems to be yet another "clamp" logic, we might take advantage of further abstracting clamping.

operator/inventory/node-discovery.go
chalabi2: Checked this. I did not zero `node.Resources` because `updateNodeInfo` only owns allocatable and capacity from `NodeStatus`. A full zero value would also drop allocated pod usage and CPU/GPU info, which are maintained by other paths. I updated the local comment to make that ownership explicit.

operator/inventory/node-discovery.go
chalabi2: Checked this. I left these clamp paths separate. `sanitizeResourceQuantity` handles invalid Kubernetes node status at ingest. `subAllocatedNLZ` handles underflow after pod-delete arithmetic and logs different context. A shared helper would hide the failure mode without reducing much code.

PR #392, opened 2026-04-22
feat: withdrawal batcher

vertex451:
Must be merged after https://github.com/akash-network/chain-sdk/pull/292

## Changes

- Added withdrawal batcher that collects withdraw messages in case of pending tx in one batch.
- Added `WithPriority` option to MsgCreateBid and MsgCloseBid

## Other callers to batch

MsgCreateBid is not suitable for batching, because failure for one bid will cancel other bids.  Same for MsgCloseBid.
No other Msgs in the provider found.

## Testing

I run a provider with:
```
    --lease-funds-monitor-interval=1m \
    --withdrawal-period=1m \
    --withdrawal-batch-max-msgs=5
```
    
And created a 10 leases.

After that I stopped a provider for one minute and run it again. Observations:
1. Pending queue goes `0→1→...→9→4→0` as expected. Zero at beginning, then first tx goes in and fires immediately, while in flight, other leases are scheduled to withdrawal, which bumps pending to 9. When tx is finished, next batch is collected, pending tx are reduced by `withdrawal-batch-max-msgs` (5).
2. Meanwhile, I run a script that created a new deployment every block, and from the logs I can confirm that priority queue is flushed before the default queue. The only case when priority queue is waiting - when the default queue tx is in progress.

### DIFF
--- a/balance_checker.go
+++ b/balance_checker.go
@@ -78,17 +78,18 @@ func newBalanceChecker(
 	bus pubsub.Bus,
 	cfg BalanceCheckerConfig,
 ) (*balanceChecker, error) {
+	bcLog := clientSession.Log().With("cmp", "balance-checker")
 	bc := &balanceChecker{
 		ctx:     ctx,
 		session: clientSession,
-		log:     clientSession.Log().With("cmp", "balance-checker"),
+		log:     bcLog,
 		bus:     bus,
 		lc:      lifecycle.New(),
 		ownAddr: accAddr,
 		aqc:     aqc,
 		leases:  make(map[mtypes.LeaseID]*leaseState),
 		cfg:     cfg,
-		batcher: newWithdrawBatcher(clientSession.Client().Tx(), withdrawTimeout, cfg.WithdrawalBatchMaxMsgs),
+		batcher: newWithdrawBatcher(clientSession.Client().Tx(), bcLog.With("cmp", "withdraw-batcher"), withdrawTimeout, cfg.WithdrawalBatchMaxMsgs),
 	}
 
 	startCh := make(chan error, 1)
--- a/balance_checker.go
+++ b/balance_checker.go
@@ -21,7 +21,6 @@ import (
 
 	"pkg.akt.dev/go/util/pubsub"
 	netutil "pkg.akt.dev/node/v2/util/network"
-	"pkg.akt.dev/node/v2/util/runner"
 	"pkg.akt.dev/node/v2/x/escrow/client/util"
 
 	"github.com/akash-network/provider/event"
@@ -43,6 +42,7 @@ const (
 type BalanceCheckerConfig struct {
 	WithdrawalPeriod        time.Duration
 	LeaseFundsCheckInterval time.Duration
+	WithdrawalBatchMaxMsgs  int
 }
 
 type leaseState struct {
@@ -60,6 +60,7 @@ type balanceChecker struct {
 	aqc     aclient.QueryClient
 	leases  map[mtypes.LeaseID]*leaseState
 	cfg     BalanceCheckerConfig
+	batcher *withdrawBatcher
 }
 
 type leaseCheckResponse struct {
@@ -87,6 +88,7 @@ func newBalanceChecker(
 		aqc:     aqc,
 		leases:  make(map[mtypes.LeaseID]*leaseState),
 		cfg:     cfg,
+		batcher: newWithdrawBatcher(clientSession.Client().Tx(), withdrawTimeout, cfg.WithdrawalBatchMaxMsgs),
 	}
 
 	startCh := make(chan error, 1)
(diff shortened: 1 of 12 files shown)


balance_checker.go
```diff
@@ -199,18 +201,6 @@ func (bc *balanceChecker) doEscrowCheck(ctx context.Context, lid mtypes.LeaseID,
 	return resp
 }
 
-func (bc *balanceChecker) startWithdraw(ctx context.Context, lid mtypes.LeaseID) error {
```
vertex451, line 202: This was moved to the withdrawal_batcher.go

withdraw_batcher_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	return f
+}
+
+func (f *batcherFixture) waitCaptured() []sdk.Msg {
+	f.t.Helper()
+	select {
+	case msgs := <-f.captured:
+		return msgs
+	case <-time.After(2 * time.Second):
```
cloud-j-luna: Make this a constant with a meaningful name

withdraw_batcher.go
```diff
@@ -0,0 +1,144 @@
...
+	copy(batch, b.pending[:n])
+	b.pending = b.pending[n:]
+	b.inFlight = true
+
+	b.log.Info("batcher: flush", "batch", n, "remaining", len(b.pending), "maxMsgs", b.maxMsgs)
+
+	go func() {
+		start := time.Now()
+		err := b.broadcast(ctx, batch)
+		b.log.Info("batcher: broadcast done", "batch", n, "dur", time.Since(start), "err", err)
```
cloud-j-luna:
```suggestion
		b.log.Info("batcher: broadcast done", "batch", n, "duration", time.Since(start), "err", err)
```

cloud-j-luna: Just a note. We have this inconsistency in the logging where error fields are named  either `err` and `error`. I believe we should keep it standard and as `err` is the widely used one we should keep it.

withdraw_batcher.go
```diff
@@ -0,0 +1,144 @@
...
+		return false
+	}
+	if len(b.pending) == 0 {
+		return false
+	}
+
+	n := len(b.pending)
+	if n > b.maxMsgs {
+		n = b.maxMsgs
+	}
```
cloud-j-luna:
```suggestion
	n := min(len(b.pending), b.maxMsgs)
```

withdraw_batcher.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// withdrawBatcher coalesces MsgWithdrawLease requests into single multi-msg
+// transactions using opportunistic in-flight batching:
+//
+//   - Idle: Flush fires a 1-msg TX immediately.
+//   - In-flight: subsequent Enqueue calls accumulate in pending.
+//   - On MarkDone: callers invoke Flush which drains up to maxMsgs from pending.
+//
+// Not safe for concurrent use. All methods except the internal broadcast
```
cloud-j-luna: Do we want to avoid concurrency altogether? We can introduce an atomic flag that triggers if there are concurrent calls. This could save us some debugging nightmares if concurrency is strictly forbidden.

cmd/provider-services/cmd/run.go
```diff
@@ -205,6 +210,10 @@ func RunCmd() *cobra.Command {
 				return fmt.Errorf(`flag "%s" value must be > "%s"`, FlagWithdrawalPeriod, FlagLeaseFundsMonitorInterval) // nolint: err113
 			}
 
+			if maxMsgs := viper.GetInt(FlagWithdrawalBatchMaxMsgs); maxMsgs < 10 || maxMsgs > 100 {
```
cloud-j-luna: make these ranges a constant that we can set and used in both the validation and the error.

cmd/provider-services/cmd/run.go
```diff
@@ -189,6 +190,10 @@ func RunCmd() *cobra.Command {
 		Short:        "run akash provider",
 		SilenceUsage: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Store logger in context before TxPersistentPreRunE so that the
+			// serialBroadcaster (created during DiscoverClient) picks it up via ctxlog.Logger(ctx).
+			fromctx.CmdSetContextValue(cmd, fromctx.CtxKeyLogc, log.NewLogger(os.Stderr))
```
cloud-j-luna, line 197: Why are we passing stderr to the logger?

cloud-j-luna (changes requested):

withdraw_batcher_test.go
vertex451: Moved to constant, named twoSecondsTestTimeout

withdraw_batcher.go
vertex451: Applied locally

vertex451: checked my diff - everywhere are used errors.
Also, I am thinking about a separate refactor PR where the error will be renamed to the err.

withdraw_batcher.go
vertex451: Applied locally

withdraw_batcher.go
vertex451: Added a protection against concurrent usage.

cmd/provider-services/cmd/run.go
vertex451: good point, done

cmd/provider-services/cmd/run.go
vertex451: I used it the way it was:
https://github.com/akash-network/provider/pull/392/changes/BASE..43783ca20d18bf404050b205bc7360fc8c3f4aae#diff-e87d9d54493149c3a09d2aff4c21a8f9d611c98662239f51100d50e21597c88aL268

Just moved upper to make sure we have logs from the chain-sdk library.

withdraw_batcher.go
```diff
@@ -0,0 +1,166 @@
...
+//
+// Not safe for concurrent use. All public methods must be called from a single
+// goroutine. Concurrent calls panic to surface developer mistakes early.
+type withdrawBatcher struct {
+	tx      aclient.TxClient
+	log     log.Logger
+	timeout time.Duration
+	maxMsgs int
+
+	inUse atomic.Int32
```
cloud-j-luna: There's an atomic.Bool type that might be more adequate here

vertex451: Makes sense, changed to bool and simplified check.

PR #396, opened 2026-05-27
feat: confidential compute

cloud-j-luna:
Adds per-pod attestation sidecar support for confidential compute workloads using **AMD SEV-SNP** and **Intel TDX**, with NVIDIA GPU confidential compute (VFIO passthrough). Tenants request CC via the existing `confidential-compute: true` placement attribute, no SDL version bump required. The provider automatically selects the appropriate Kata runtime class (`kata-qemu-snp`, `kata-qemu-nvidia-gpu-snp`, `kata-qemu-tdx`, `kata-qemu-nvidia-gpu-tdx`), applies CC-specific node selectors, switches GPU resources to `nvidia.com/pgpu` for VFIO, and injects an attestation sidecar into the pod via a fail-closed mutating admission webhook. The sidecar serves an HTTP endpoint inside the Kata VM that collects raw hardware-signed attestation reports (`configfs-tsm` or `/dev/sev-guest`) and returns them as is, **the provider never produces, inspects or signs attestation evidence**. Tenants verify reports independently against trusted parties.

New provider REST endpoints (GET `/attestation/directory` and POST `/lease/.../attestation/quote`) and a lease-attestation CLI command enable the full attestation flow. Tenants can opt out of sidecar injection via `params.attestation.enabled: false` in the SDL.

Local development is supported with `CONFIDENTIAL_COMPUTE=true` a single variable that configures Kind with mock runtime handlers, node labels, and a mock TEE provider that produces synthetic reports with correct nonce echo for end-to-end testing without hardware.


cluster/types/v1beta3/clients/inventory/inventory.go
```diff
@@ -406,15 +425,12 @@ func tryAdjustGPU(rp *inventoryV1.GPU, res *rtypes.GPU, sparams *crd.SchedulerPa
 				return false
 			}
 
-			// sParamsEnsureGPU(sparams)
 			sparams.Resources.GPU.Vendor = vendor
 			sparams.Resources.GPU.Model = info.Name
```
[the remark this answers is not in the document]
cloud-j-luna: This was already addressed locally

go.mod
```diff
@@ -98,6 +99,9 @@ replace (
 	golang.org/x/exp => golang.org/x/exp v0.0.0-20230711153332-06a737ee72cb
 	// stick with compatible version of rapid in v0.47.x line
 	pgregory.net/rapid => pgregory.net/rapid v0.5.5
+
+	// local node development (v2 tag with LeaseStartReclaim stub)
+	pkg.akt.dev/node/v2 => ../node
 )
```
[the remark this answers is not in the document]
cloud-j-luna: dependency on chain-sdk will fix later

cluster/kube/builder/workload.go
```diff
@@ -305,6 +422,38 @@ func (b *Workload) affinity() *corev1.Affinity {
...
 
+	if params != nil && IsConfidentialComputeRuntimeClass(params.RuntimeClass) {
+		selectors = append(selectors, corev1.NodeSelectorRequirement{
+			Key:      "katacontainers.io/kata-runtime",
+			Operator: corev1.NodeSelectorOpIn,
+			Values:   []string{"true"},
+		})
+
+		// TEE-specific node labels
+		if IsSNPRuntimeClass(params.RuntimeClass) {
```
cloud-j-luna: @troian runtime selection is based on node labels. To correct what I mentioned early about believing the inventory operator was handling it, I mistakenly confused it with the inventory client actually setting the runtime and it getting propagated here. Was this what you were asking? SNP and TDX are obviously mutually exclusive but I'm not sure a switch statement here would make that much readability difference.

sidecar/attestation/tee/tee.go
```diff
@@ -0,0 +1,127 @@
...
+	Report      []byte `json:"report"`
+}
+
+// QuoteResult holds the raw hardware-signed attestation evidence.
+type QuoteResult struct {
+	Report     []byte            // Raw attestation report (SNP ~1184 bytes, TDX 1024 bytes)
+	CertChain  []byte            // Cert chain (may be empty — fetching is tenant-side)
+	AuxBlob    []byte            // Empty on NVIDIA-patched kernel
+	GPUReports []GPUDeviceReport // Per-GPU attestation reports (nil when no GPU CC)
+}
```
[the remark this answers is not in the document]
cloud-j-luna: The NVML C helper controls the framing. This is a non-issue today

cmd/provider-services/cmd/run.go
```diff
@@ -813,6 +826,128 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
...
+			Log:          webhookLog,
+		})
+
+		// Register the MutatingWebhookConfiguration so the K8s API server
+		// sends pod CREATE requests to our webhook for sidecar injection.
+		// The CA bundle must be PEM-encoded. For self-signed certs, it's the leaf cert itself.
+		caBundle := pem.EncodeToMemory(&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: tlsCert.Certificate[0],
+		})
```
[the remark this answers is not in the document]
cloud-j-luna: This is anon issue right now

cmd/provider-services/cmd/run.go
```diff
@@ -813,6 +826,128 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
...
+	// Start attestation webhook server if enabled
+	if viper.GetBool(FlagAttestationWebhookEnabled) {
+		sidecarImage := viper.GetString(FlagAttestationSidecarImage)
+		if sidecarImage == "" {
+			return fmt.Errorf("%w: %s is required when %s is enabled",
+				errInvalidConfig, FlagAttestationSidecarImage, FlagAttestationWebhookEnabled)
+		}
+		webhookPort := viper.GetInt(FlagAttestationWebhookPort)
+		webhookAddr := fmt.Sprintf(":%d", webhookPort)
+
```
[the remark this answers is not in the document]
cloud-j-luna: The net.Listen call on that address will return a clear error

_run/kube/cc-containerd-setup.sh
troian, line 1: lets keep naming for the files consistent.
put `cc` as suffix, e.g.

setup-cc.yaml
deployment-gpu-cc.yaml

remove `cc-conteinerd-setup.sh` and put it's content into `script/setup-kube.sh` as command
if this file is intended to run on cluster, then it should be deployed via kustomize in `_docs/kustomize`

_run/kube/Makefile
```diff
@@ -52,6 +73,40 @@ provider-lease-ping:
 hostname-operator:
 	$(PROVIDER_SERVICES) hostname-operator
 
+# Confidential compute cluster setup: creates RuntimeClass objects and labels
+# the Kind node with TEE capability labels so the provider can schedule CC
+# workloads. The pods won't actually run with Kata (Kind doesn't have it),
+# but the webhook, annotations, and sidecar injection can be verified.
+.PHONY: kube-setup-cc
```
troian: there is no need for separate make run command.

setup should be using same `kube-setup` target which is being configured using env variables.
if state needs to be saved there is `.cache` for temporary files as well as you can query kube cluster for details

same applies for `kube-teardown`

cluster/kube/builder/workload.go
```diff
@@ -17,12 +17,59 @@ import (
...
-	GPUVendorNvidia   = "nvidia"
-	GPUVendorAMD      = "amd"
+	ResourceGPUNvidia     = corev1.ResourceName("nvidia.com/gpu")
+	ResourceGPUNvidiaPGPU = corev1.ResourceName("nvidia.com/pgpu")
+	ResourceGPUAMD        = corev1.ResourceName("amd.com/gpu")
+	GPUVendorNvidia       = "nvidia"
+	GPUVendorAMD          = "amd"
 )
 
+// IsConfidentialComputeRuntimeClass returns true if the given runtime class
```
troian: there are way to many functions for the same thing
1. functions naming is a bit cumbersome
2. I feel like single `IsRuntimeClass(rc string, opts ...Option)` handling all the cases will do the job. plus code will look consistent for all callers. Options will allow filtering down specific classes. e.g `WithCC` will check if provided runtime class CC related

cluster/kube/builder/workload.go
```diff
@@ -94,25 +141,50 @@ func (b *Workload) container() corev1.Container {
...
-		SecurityContext: &corev1.SecurityContext{
-			RunAsNonRoot:             &falseValue,
-			Privileged:               &falseValue,
-			AllowPrivilegeEscalation: &falseValue,
-		},
+	}
+
+	sidecarActive := sparams != nil &&
+		IsConfidentialComputeRuntimeClass(sparams.RuntimeClass) &&
+		!sparams.AttestationDisabled
```
troian, line 132: this should be `sparams member instead`
```
function (s *SchedulerParams) IsCCSidecarActive() bool {
    return s != nil && !s.AttestationDisabled && IsRuntimeClass(s.RuntimeClass, WithCC())
}
```

troian (changes requested):

cluster/kube/builder/workload.go
cloud-j-luna: This function is used also by the sidecar injector as well. Moving it as a member of `sparams` would require me to duplicate that specific check or have it as a wrapper which adds another layer. This function would still be required in the end

cluster/kube/builder/workload.go
cloud-j-luna: Totally agree!

_run/kube/Makefile
cloud-j-luna: The decision to have it as a standalone target comes from the fact that the CC setup code is substantial enough to warrant being a named internal target. I thought having it all in the same target would make it more cumbersome

cluster/kube/client_attestation.go
```diff
@@ -0,0 +1,145 @@
...
+// Returns TEEPlatformTDX if any node has the TDX label, TEEPlatformSNP for SNP,
+// or TEEPlatformNone if no CC-capable nodes are found.
+func (c *client) DetectTEEPlatform(ctx context.Context) ctypes.TEEPlatform {
+	nodes, err := c.kc.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return ctypes.TEEPlatformNone
+	}
+
+	for _, node := range nodes.Items {
+		if node.Labels[intelTDXLabelKey] == "true" {
```
troian: there is const in builder for both `true` and `false`.

troian (approved):

PR #397, opened 2026-05-27
feat(aep-86): add provider signer abstraction

chalabi2:
## Summary
- Add a narrow provider signer interface backed by the existing provider client context and keyring.
- Thread the signer into provider service startup so later AEP-86 inventory and snapshot components can sign provider-owned payloads and transactions.
- Share Bearer token parsing between REST and gRPC gateway paths so both enforce the same auth shape.

## Scope
- AEP-86 L-2 provider signer foundation.
- This does not add inventory snapshots, snapshot hash posting, bid filtering, or node verification-module logic.
- Follow-up PRs will build L-3 inventory snapshots and L-10 snapshot posting on top of this signer path.

## Linear
Refs AKT-287

### DIFF
--- a/gateway/rest/auth.go
+++ b/gateway/rest/auth.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"log"
 	"net/http"
-	"strings"
 
 	gcontext "github.com/gorilla/context"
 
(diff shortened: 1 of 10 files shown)


gateway/rest/auth.go
```diff
@@ -16,17 +15,7 @@ import (
...
-		return "", nil // No error, just no JWT.
-	}
-
-	authHeaderParts := strings.Fields(authHeader)
-	if len(authHeaderParts) != 2 || strings.ToLower(authHeaderParts[0]) != "bearer" {
-		return "", httperror.ErrInvalidAuthHeader
-	}
-
-	return authHeaderParts[1], nil
+	return gwutils.AuthHeaderToken(r.Header.Values("Authorization"))
```
troian, line 18: lets delete `AuthHeaderTokenExtractor`, callers should just use `gwutils.AuthHeaderToken`

troian (changes requested):

PR #315, opened 2025-08-22
feat: add provider-url support to lease commands for on-chain independence

cloud-j-luna:
Adds provider URL support across lease-related commands, introduces a new provider-url flag and helper-based client construction, migrates provider client creation to option-based WithQueryClient, updates tests accordingly, adjusts Make targets to include provider-url variants, and bumps akash-api to v0.0.84.

[API PR](https://github.com/akash-network/akash-api/pull/229)

## How to test

```bash
$ make provider-lease-shell
provider-services lease-shell \
        --dseq      "1"        \
        --gseq      "1"        \
        --oseq      "1"        \
        --from      "main"    \
        --provider  "akash19mp92vqx6vu44uldmumtyjceladnmzegk94uf5" \
        --auth-type=jwt \
        web \
        date
Wed Sep 24 10:26:51 UTC 2025

$ make provider-url-lease-shell
provider-services lease-shell \
        --dseq      "1"        \
        --gseq      "1"        \
        --oseq      "1"        \
        --from      "main"    \
        --provider  "akash19mp92vqx6vu44uldmumtyjceladnmzegk94uf5" \
        --provider-url "https://localhost:8443" \
        --auth-type=jwt \
        web \
        date
Wed Sep 24 10:26:56 UTC 2025

##### NODE GOES OFFLINE ######

$ make provider-url-lease-shell
provider-services lease-shell \
        --dseq      "1"        \
        --gseq      "1"        \
        --oseq      "1"        \
        --from      "main"    \
        --provider  "akash19mp92vqx6vu44uldmumtyjceladnmzegk94uf5" \
        --provider-url "https://localhost:8443" \
        --auth-type=jwt \
        web \
        date
Wed Sep 24 10:27:01 UTC 2025

$ make provider-lease-shell
provider-services lease-shell \
        --dseq      "1"        \
        --gseq      "1"        \
        --oseq      "1"        \
        --from      "main"    \
        --provider  "akash19mp92vqx6vu44uldmumtyjceladnmzegk94uf5" \
        --auth-type=jwt \
        web \
        date
Error: post failed: Post "http://localhost:26657": dial tcp [::1]:26657: connect: connection refused
make: *** [../common-commands.mk:346: provider-lease-shell] Error 1
```

PR #400, opened 2026-06-03
feat: GPU interconnect multinode support

Zblocker64:
Provider side of the GPU interconnect feature. Multi-node GPU workloads (NCCL over InfiniBand or RoCE) can now land on Akash providers. Consumes the released chain-sdk `pkg.akt.dev/go v0.4.1` + `pkg.akt.dev/go/sdl v0.4.1` (interconnect SDL surface merged upstream in chain-sdk #315).

Single squashed commit, rebased onto main (includes the confidential compute base).

## SDL surface (chain-sdk v0.4.1)

Tenants opt in per service with `gpu.attributes.interconnect: []` (implicit shared `auto` group) or `interconnect: { group: <name> }` (explicit groups). Works with any `gpu.units ≥ 1`. Placement must require `capabilities/gpu-interconnect: "true"`, with an optional `capabilities/gpu-interconnect/fabric/{infiniband,roce}` pin. On-chain signal is the single `interconnect/group` GPU attribute.

## What's in here

- **Inventory operator** — discovers fabric (IB/RoCE) and NCCL HCA device-name families (`nccl_hca_prefixes`, array — mixed-vendor hosts publish every family) from `/sys/class/infiniband`, and the interconnect kubelet extended resource via **configurable glob patterns** (`interconnect.resource_patterns`, default `rdma/rdma_shared_device_*` — non-Mellanox plugins supported). Stale capability fields clear when the device plugin resource disappears.
- **Reservation Adjust** — one HCA handle per GPU (1:1 with `gpu.units`); per-group node-claim tracking so peers of one interconnect group always land on distinct nodes (the bid refuses what the kube scheduler couldn't place). Composes with confidential compute (`tryAdjust` carries TEE type/platform alongside the interconnect gates).
- **Workload builder** — requests the discovered extended resource (`request==limit`); injects NCCL env: `NCCL_IB_DISABLE=0`, `NCCL_IB_HCA` (comma-joined families), and `NCCL_IB_GID_INDEX=3` on RoCE. SDL-set values always win. Per-group pod label + hard anti-affinity on `kubernetes.io/hostname`.
- **CRD (v2beta2)** — `SchedulerResourceInterconnect{enabled, units, resource_name, fabric, nccl_hca_prefixes}` + `Service.interconnect_group`, with schema additions so the API server doesn't prune them.

## Linear

AKT-407..412 (P-1..6), AKT-436..439 (review fixes), AKT-443 (group-aware Adjust), AKT-492 (v0.4.1 SDL surface), AKT-493 (configurable resource patterns), AKT-494 (RoCE GID index).

## Test plan

- [x] Full `go test ./...` suite
- [x] End-to-end on a 2-node A100/IB cluster: bid → lease → manifest → pods on distinct nodes with `rdma/rdma_shared_device_ib` allocated + NCCL env injected
- [ ] E2E re-run on v0.4.1 nested SDL form
- [ ] E2E on RoCE
- [ ] E2E multi-group


operator/inventory/node-discovery.go
```diff
@@ -905,6 +1104,52 @@ func (dp *nodeDiscovery) parseCPUInfo(ctx context.Context) v1.CPUInfoS {
...
+	if dev.Subclass != nil {
+		subclassID = dev.Subclass.ID
+	}
+
+	// Compute GPUs are PCI class 03 subclass 02 ("3D controller"). Headless
+	// A100/H100 boxes don't expose them as display controllers (03/00) and
+	// don't have DRM nodes, so this is the only reliable filter. Skipping
+	// 03/00 keeps the BMC VGA out, and skipping anything outside class 03
+	// keeps NVSwitch bridges (06/80) and similar out.
+	if classID != "03" || subclassID != "02" {
```
chalabi2: Potential filtering regression? It looks like this drops consumer GPUs. This requires PCI class 03/subclass 02 but consumer cards in our registry (RTX 4090/3090, A6000, etc.)
enumerate as 03/00 (VGA compatible controller).

go.mod
```diff
@@ -51,9 +51,9 @@ require (
 	k8s.io/client-go v0.34.1
 	k8s.io/code-generator v0.34.1
 	k8s.io/kubectl v0.33.3
-	pkg.akt.dev/go v0.2.10
+	pkg.akt.dev/go v0.2.13-rc1
```
chalabi2: Does it matter that we are pinning RC's here?

cluster/kube/operators/clients/inventory/inventory.go
```diff
@@ -43,14 +43,31 @@ func (inv *inventory) Dup() ctypes.Inventory {
...
 		return nil, false, true
 	}
 
+	// tryAdjustRDMA must run BEFORE tryAdjustGPU because tryAdjustGPU
+	// replaces res.GPU.Attributes with a single synthesized vendor entry
+	// (clobbering the `rdma: true` opt-in). ResourceRequiresRDMA would
+	// then return false and the reservation would succeed as a non-RDMA
+	// bid — wrong: the workload builder would skip the RDMA resource
+	// request and NCCL env injection.
+	if !tryAdjustRDMA(&nd.Resources.RDMA, nd.Capabilities, res, sparams, requiredFabric) {
```
chalabi2: does `tryAdjustGPU` wipe the rdma: true flag? (line 208). Looks like any RDMA group that spans more than one node loses RDMA on the second node and the bid gets rejected.

cluster/kube/builder/workload.go
```diff
@@ -335,6 +350,31 @@ func (b *Workload) affinity() *corev1.Affinity {
...
+	// `gpu.attributes.rdma_group` must have its replicas (and any other
+	// services that share the same group label) land on distinct nodes,
+	// because we allocate one RDMA HCA per GPU per node and peers in a
+	// group expect a 1:1 GPU:HCA fanout. Requirement (not preference) so
+	// the kube scheduler hard-rejects co-location.
+	//
+	// Scoping the LabelSelector to the same deployment namespace is
+	// implicit — pod affinity is namespace-scoped by default — so two
+	// tenants who happen to both pick `rdma_group: pair0` cannot collide.
+	if rg := service.RDMAGroup; rg != "" {
```
chalabi2: I think this forces group peers onto separate nodes, but the bid step doesn't know about rdma_group or that it can fit several peers on one node.

chalabi2 (changes requested):

go.mod
Zblocker64: We will need to bump to non RCs once that PR is merged

cluster/kube/builder/workload.go
Zblocker64: You're right. The chain SDK serializes `rdma_group` only on the off-chain manifest's `Service.RDMAGroup` field — it doesn't survive into the on-chain Order's `GPU.Attributes` (confirmed by bid logs during e2e). So `tryAdjust` can't enforce per-group node separation at fit time. The hard `PodAntiAffinity` here is the only check, and it fires too late — after the bid has already accepted.

Two failure modes this creates:

(a) `count: N` of one service in one group on a provider with fewer than N distinct GPU-capable nodes → bid succeeds greedily, K8s leaves the surplus pods Pending forever.
(b) Multi-service group with enough total GPU/HCA capacity on a single node → bid greedily co-locates peers, K8s leaves the second peer Pending.

Our 2-node POC dodged both by capacity coincidence (1 head + 1 worker, count=1 each, exactly 8 GPUs per node).

I've added a comment block above this clause in `bf774e8` flagging the gap and the two failure modes, and filed AKT-443 — "Bid engine: group-aware Adjust for rdma_group peers" — for the real fix. The ticket has design notes for three options (carry `rdma_group` into the order as a synthetic attribute, plumb the off-chain manifest into `Adjust`, or switch to soft anti-affinity — only the first two are real fixes).

Keeping this PR scoped to what we've proven works (single-replica per service, single-group multi-node). The real fix lands in AKT-443.

Zblocker64: Implemented the fix for this

chalabi2 (dismissed):
lgtm - lets remember to bump the chain-sdk version when its merged and released

Zblocker64:
@coderabbitai full review

troian (approved):

PR #404, opened 2026-06-17
fix: copy GPU attributes when converting resources

miloquinn:
## Summary
- Copy GPU attributes with `Dup()` in `resourcesFromAkash`, matching CPU, memory, and storage handling.
- Add a regression test that mutates converted GPU attributes and verifies the source Akash resource is unchanged.

## Testing
- `GOTOOLCHAIN=auto go test ./pkg/apis/akash.network/v2beta2`


troian:
@miloquinn thanks. do you have some context on why this change is necessary. Is there a bug

Zblocker64:
closing due to staleness, @miloquinn feel free to reopen if still needed

PR #405, opened 2026-06-17
chore(docker): include grpcurl in provider image

### LINKED ISSUE #317: Include grpcurl in Provider Image
Feature request:

Include the insall of `grpcurl` in provider image to allow testing within the inventory operator and automation of pod restarts when conditions warrant.

Example use case:

Liveness probe for the inventory operator would restart pod when GPU allocated is greater than capacity/allocatable.

neyy91:
## Summary

Resolves akash-network/support#317.

The inventory operator exposes a gRPC API (`akash.inventory.v1.ClusterRPC`). Bundling `grpcurl` in the provider image makes it possible to probe that API from inside the container - for example, a liveness probe that restarts the pod when allocated resources exceed allocatable.

## Change

`grpcurl` is not available via apt, so the Dockerfile installs a pinned release:

- Pinned to `v1.9.3`.
- Works for both **amd64** and **arm64** (the image is built for both), selecting the right asset via `dpkg --print-architecture`.
- The downloaded tarball is **sha256-verified** against the published checksum before extraction.
- Only the `grpcurl` binary is installed; the tarball is removed afterwards.
- `grpcurl -version` runs at the end so the build fails fast if anything is wrong.

## Testing

Built the image locally (amd64) and confirmed the install step downloads, passes the checksum check, extracts the binary, and `grpcurl -version` reports `v1.9.3`. The arm64 asset uses the official published checksum.

refs: akash-network/support#317

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,28 @@ RUN \
 
 ENV DEBIAN_FRONTEND=""
 
+# grpcurl is used to probe the inventory operator's gRPC API (e.g. from a
+# liveness probe). It is not packaged in apt, so install a pinned release.
+# The version and per-arch checksums are supplied as build args (sourced from
+# .env via the goreleaser build), so bumping the release does not require
+# editing this Dockerfile.
+ARG GRPCURL_VERSION
+ARG GRPCURL_SHA256_AMD64
+ARG GRPCURL_SHA256_ARM64
+RUN set -eux; \
+    case "$(dpkg --print-architecture)" in \
+      amd64) arch=x86_64; sha="${GRPCURL_SHA256_AMD64}" ;; \
+      arm64) arch=arm64;  sha="${GRPCURL_SHA256_ARM64}" ;; \
+      *) echo "unsupported architecture: $(dpkg --print-architecture)" >&2; exit 1 ;; \
+    esac; \
+    curl -fL --retry 5 --retry-delay 2 --retry-connrefused --connect-timeout 10 --max-time 120 \
+      -o /tmp/grpcurl.tar.gz \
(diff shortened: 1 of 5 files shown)


chalabi2 (dismissed):
lgtm

Dockerfile
```diff
@@ -20,6 +20,22 @@ RUN \
 
 ENV DEBIAN_FRONTEND=""
 
+# grpcurl is used to probe the inventory operator's gRPC API (e.g. from a
+# liveness probe). It is not packaged in apt, so install a pinned release.
+ENV GRPCURL_VERSION=1.9.3
```
troian: `ARG GRPCURL_VERSION=1.9.3`

Dockerfile
```diff
@@ -20,6 +20,22 @@ RUN \
...
+RUN set -eux; \
+    case "$(dpkg --print-architecture)" in \
```
troian, line 32: lets hardcode checksums into RUN, there is no need to split them as well, `sha256sum` supports checksums format from github releases

troian (changes requested):

Dockerfile
neyy91: switched GRPCURL_VERSION to ARG (build-time only). On the checksums - to make sure I match what you have in mind: do you mean fetching grpcurl's upstream checksums.txt and verifying the downloaded archive against it, or just hardcoding the two <sha>  <filename> lines verbatim and letting sha256sum -c match by filename? Dockerfile RUN can't use a shell heredoc, so I'd do it via printf. Any option works for me-just let me know which one you prefer.

troian: checksums can be hardcoded in the .env file, so that way we don't have to touch dockerfile should we need to bump the version

troian (approved):

PR #406, opened 2026-06-19
fix(inventory): exclude tainted nodes from the inventory

neyy91:
### What

Node discovery (`operator/inventory/node-discovery.go`) only excluded nodes that were not-ready, cordoned (`Unschedulable`), or in the operator exclude list. Nodes carrying a `NoSchedule`/`NoExecute` taint were still reported as available capacity.

This adds `nodeHasBlockingTaint` and treats any `NoSchedule`/`NoExecute` taint as a reason to keep the node out of the inventory. `PreferNoSchedule` is a soft preference and is left alone.

### Why

Closes the behaviour reported in akash-network/support#253: providers with tainted nodes (e.g. `node-role.kubernetes.io/control-plane`, `CriticalAddonsOnly`) kept bidding using that capacity once the rest of the cluster was full. Akash pods carry no custom tolerations, so the pod can never schedule onto a tainted node - the deployment fails `FailedScheduling` and the provider closes it minutes later. Reported in the wild on multiple providers and on a support call.

### Testing

Added `TestGenerateLabels_Taints`: a ready, untainted node is marked akash-managed; `NoSchedule` and `NoExecute` tainted nodes are excluded; `PreferNoSchedule` is not. Fails without the fix.

refs: akash-network/support#253

### DIFF
--- a/operator/inventory/node-discovery.go
+++ b/operator/inventory/node-discovery.go
@@ -1049,24 +1049,21 @@ func removeManagedLabels(in map[string]string) map[string]string {
 	return out
 }
 
-// nodeHasBlockingTaint reports whether any taint would prevent Akash workloads
-// from being scheduled onto the node. Akash pods carry no custom tolerations,
-// so a NoSchedule or NoExecute taint means the node cannot accept workloads and
-// its capacity must be kept out of the inventory; otherwise the provider bids
-// on capacity it can never use and the deployment fails to schedule.
+// isNodeReady reports whether the node can actually accept Akash workloads. It
+// must be both in the Ready condition and free of any scheduling-blocking
+// taint: Akash pods carry no custom tolerations, so a NoSchedule/NoExecute
+// taint (e.g. control-plane, CriticalAddonsOnly) means the node can never host
+// a workload and its capacity must stay out of the inventory; otherwise the
+// provider bids on capacity it cannot use and the deployment fails to schedule.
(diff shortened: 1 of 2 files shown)


operator/inventory/node-discovery.go
```diff
@@ -824,7 +840,7 @@ func generateLabels(cfg Config, knode *corev1.Node, node v1.Node, sc storageClas
 	sort.Strings(presentSc)
 	adjConfig.FilterOutStorageClasses(presentSc)
 
-	isExcluded := !isNodeReady(knode.Status.Conditions) || knode.Spec.Unschedulable || adjConfig.Exclude.IsNodeExcluded(knode.Name)
+	isExcluded := !isNodeReady(knode.Status.Conditions) || knode.Spec.Unschedulable || nodeHasBlockingTaint(knode.Spec.Taints) || adjConfig.Exclude.IsNodeExcluded(knode.Name)
```
troian: lets keep content of `nodeHasBlockingTaint` within `isNodeReady`

troian (changes requested):

operator/inventory/node-discovery.go
neyy91: Done. folded the taint check into isNodeReady, which now takes the node and treats a NoSchedule/NoExecute taint as not-ready. Dropped the separate helper.

troian (approved):

PR #409, opened 2026-06-21
fix(bidengine): increment reservation counter metrics

neyy91:
### What

Four `reservationCounter.WithLabelValues(...)` calls in `bidengine/order.go` were missing a terminal `.Inc()`, so the returned counter was discarded and the metric was never recorded.

### Why

Every other counter in this file (`orderCompleteCounter`, `shouldBidCounter`, `bidCounter`) calls `.Inc()`; these four (reservation open/close × success/fail) were the exception. As a result the reservation metrics stayed flat at zero, so operators had no visibility into reservation success/failure rates.

### Testing

Added `Test_ReservationCounterIncrements`: runs a full reserve + unreserve cycle and asserts the open-success and close-success counters each advance by one. Fails without the `.Inc()` calls. (`go mod tidy` registered `prometheus/.../testutil`'s indirect dep `kylelemons/godebug`; go.sum unchanged.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -280,6 +280,7 @@ require (
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/kr/text v0.2.0 // indirect
 	github.com/kube-object-storage/lib-bucket-provisioner v0.0.0-20221122204822-d1a8c34382f1 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/lib/pq v1.10.9 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -280,6 +280,7 @@ require (
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/kr/text v0.2.0 // indirect
 	github.com/kube-object-storage/lib-bucket-provisioner v0.0.0-20221122204822-d1a8c34382f1 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
troian, line 283: ✂️

troian (approved):

PR #411, opened 2026-06-22
fix: increase proxy-buffer-size default to 16k

cloud-j-luna:


troian (changes requested):
annotation should be configurable at least via env variable, even tho this is temporary

cluster/kube/gateway/nginx_gateway.go
```diff
@@ -146,6 +148,12 @@ func (n *nginxGateway) BuildAnnotations(directive chostname.ConnectToDeploymentD
 		annotations["nginx.org/proxy-next-upstream"] = strBuilder.String()
 	}
 
+	// TODO(temporary): read proxy-buffer-size from viper directly to avoid
+	// plumbing through GatewayConfig; move to GatewayConfig when more settings need it.
+	if v := viper.GetString("proxy-buffer-size"); v != "" {
```
troian: should use `FlagProxyBufferSize`

operator/hostname/cmd.go
```diff
@@ -116,6 +116,11 @@ func addGatewayApiFlags(cmd *cobra.Command) {
 	if err := viper.BindPFlag("gateway-provider", cmd.Flags().Lookup("gateway-provider")); err != nil {
 		panic(err)
 	}
+
+	cmd.Flags().String("proxy-buffer-size", "16k", "NGINX proxy buffer size for upstream response headers")
```
troian: `FlagProxyBufferSize`

Dockerfile.debug
```diff
@@ -6,7 +6,7 @@ COPY provider-services /usr/bin/
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN \
-    go install github.com/go-delve/delve/cmd/dlv@latest \
+    go install github.com/go-delve/delve/cmd/dlv@v1.26.3 \
```
troian: lets not bind version here

if it needs to be affixed then pass it as build arg value of which is based on current GOTOOLCHAIN set by direnv

troian (dismissed):

Dockerfile.debug
cloud-j-luna: Kept this as latest and now there is an ARG to set the base go image for the debug dockerfile

chalabi2 (approved):

PR #412, opened 2026-06-22
fix(builder): key GPU interface node selector by interface, not memory size

neyy91:
### What

The GPU `interface` node-selector requirement in `cluster/kube/builder/workload.go` was built with `gpu.MemorySize` in the label key instead of `gpu.Interface`:

```go
if gpu.Interface != "" {
    Key: fmt.Sprintf("%s.interface.%s", key, gpu.MemorySize) // -> gpu.Interface
}
```

### Why

The inventory operator publishes the node label as `<gpu>.interface.<interface>` (`operator/inventory/node-discovery.go`, via `info.Interface`). Because the builder keyed the selector off the memory size instead, a workload requesting a specific GPU interface (e.g. `pcie`) produced a node-affinity requirement that can never match a real node label — so the pod fails to schedule.

### Testing

Added `Test_nodeSelectorsFromResources_GPUInterface`: a GPU with `memory_size=80Gi`, `interface=pcie` must yield a selector keyed `...interface.pcie`, and must NOT yield `...interface.80Gi`. Fails before the fix

### DIFF
--- a/cluster/kube/builder/workload.go
+++ b/cluster/kube/builder/workload.go
@@ -537,7 +537,7 @@ func nodeSelectorsFromResources(res *crd.SchedulerResources) []corev1.NodeSelect
 
 		if gpu.Interface != "" {
 			selectors = append(selectors, corev1.NodeSelectorRequirement{
-				Key:      fmt.Sprintf("%s.interface.%s", key, gpu.MemorySize),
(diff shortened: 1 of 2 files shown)


troian:
@chainzero test if breaks existing workflows

troian (approved):

PR #413, opened 2026-06-23
fix(inventory): preserve raw negative state

### LINKED ISSUE #429: Inventory operator reports absurd GPU counts
**Context**
In some cases, the inventory operator reports millions of available GPUs on a host that physically has only a few (e.g. 8).
Likely causes (to confirm in code):
- Integer underflow/overflow when computing available = total - used.
- Bad or transient values from the NVIDIA device plugin / DRA (uninitialized, negative, or huge) not being validated.

**Proposed fix**
- Add strict validation around GPU counts: enforce 0 <= used <= total and 0 <= available <= total; if out of range, clamp to safe values and log a warning.
- Use safe integer types for intermediate calculations to avoid underflow/overflow, then validate before storing.
- Add a sanity cap on per‑node GPU totals (e.g. ignore/report error if total > 128). Or use another type to hold bigger values.

chalabi2:
## Summary
Replaces https://github.com/akash-network/provider/pull/379 with a narrower fix for https://github.com/akash-network/support/issues/429.

This keeps raw operator inventory intact so negative resource accounting remains visible, while bid, metrics, and provider status consume a sanitized non-negative view.

It also logs forensic data when accounting goes negative: pod/PV identity, `inventory_before`, and the applied change. Rancher storage accounting now uses signed resource quantities instead of `uint64` arithmetic so underflows stay visible instead of becoming huge positive values.

Fixes akash-network/support#429.

## Validation
- `GOWORK=off go test ./operator/inventory -count=1`
- `GOWORK=off go test ./cluster/... ./operator/inventory -count=1`
- `git diff --check`

### DIFF
--- a/cluster/inventory.go
+++ b/cluster/inventory.go
@@ -47,6 +47,7 @@ var (
 	errInventoryReservation     = errors.New("inventory error")
 	errNoLeasedIPsAvailable     = fmt.Errorf("%w: no leased IPs available", errInventoryReservation)
 	errInsufficientIPs          = fmt.Errorf("%w: insufficient number of IPs", errInventoryReservation)
+	inventoryStatusQuantityZero = resource.NewQuantity(0, resource.DecimalSI)
 )
 
 var (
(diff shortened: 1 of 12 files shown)


chalabi2:
Closing this PR because the branch was accidentally stacked on the old PR 379 branch instead of starting from main. I am rebuilding the replacement from a clean origin/main branch.

PR #415, opened 2026-06-24
feat(bidengine): support wildcard gpu multi-bids

chalabi2:
## Summary
Adds provider multi-bid support for wildcard GPU model orders.

The bid engine now increments `BSeq`, reserves per bid, submits per-bid resource offers, and frees losing reservations when one local bid wins.

## Tests
- `GOWORK=off go test ./...`

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -6,7 +6,7 @@ COPY provider-services /usr/bin/
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN \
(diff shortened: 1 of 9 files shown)


chalabi2:
addressed the review comments:

- removed the redundant embedded `Resources` selector in wildcard GPU detection
- drained `storedGroupCh` on early exits
- only mark bids placed after broadcast success
- prevent late reservation results from resurrecting externally closed bids
- confirm leased IP reservations by exact `BidID`
- fixed existing-bid tests to match the open-bid query filter

PR #424, opened 2026-07-30
feat(kube): attach rail networks to RoCE interconnect workloads

Zblocker64:
## Problem

GPU interconnect (#400) validated end-to-end on InfiniBand but not on RoCE. On a RoCE cluster, interconnect pods received the RDMA device, correct scheduling, and NCCL env — but `ib_write_bw` between pods failed at `Failed to modify QP to RTR`.

Root cause: **RoCEv2 is IP-addressed.** QP setup resolves the remote rail IP through the pod's own network namespace, so the RDMA verbs device alone is not enough — the pod needs the rail netdevs (and the RoCEv2 GIDs they register) in its netns. InfiniBand is LID-addressed and needs none of this, which is exactly why IB passed while RoCE failed.

## Fix

**Attach the rail networks to RoCE interconnect pods.**

- At deploy time, when a deployment carries a RoCE-pinned interconnect reservation, the kube client lists multus `NetworkAttachmentDefinitions` in a dedicated rails namespace (`--interconnect-roce-networks-namespace`, default `akash-rails`) and applies them all — sorted, as `namespace/name` refs — via the `k8s.v1.cni.cncf.io/networks` pod annotation.
- Operators create one NAD per rail in that namespace (MacvlanNetwork + nv-ipam IPPool via the NVIDIA Network Operator); adding a rail later requires no provider restart or config change. With the default namespace there is nothing to configure on the provider.
- **Drops the hardcoded `NCCL_IB_GID_INDEX=3` injection for RoCE.** With rail netdevs attached, the pod's RoCEv2 GID registers at a pod-specific index and host GID entries are not visible from the pod netns — any fixed value is wrong. NCCL auto-selects the correct GID from the pod's own table. SDL-supplied overrides still pass through.

IB providers are unaffected: the NAD listing only runs for `fabric == roce` deployments, the annotation is gated on the same check, and IB never received the GID env var.

## Validation

4-node Dell 8× H200 SXM cluster, 8× ConnectX-7 rails per node (400G each, Ethernet mode / RoCEv2), NVIDIA Network Operator 26.1.1, nv-ipam auto-assigned pod IPs. Real marketplace lease (not a hand-built pod):

- Provider auto-attached all 8 rails: pods came up with `net1`–`net8` + rail IPs, annotation `akash-rails/rail0,…,rail7` stamped with zero operator config
- NCCL env: `NCCL_IB_DISABLE=0`, `NCCL_IB_HCA=mlx5`, no GID index
- Services on distinct nodes (interconnect anti-affinity)
- Pod-to-pod `ib_write_bw` over a rail: **391.37 Gb/s** — line rate, identical to the raw host-to-host baseline
- 8-rail concurrent aggregate (pre-provider harness, same pod shape): **2.09+ Tb/s**

Before the fix, the same lease shape failed QP RTR with only the cluster CNI interface in the pod.

## Notes

- Provider RBAC needs `list` on `network-attachment-definitions.k8s.cni.cncf.io` (helm-charts follow-up; broad SAs already have it)
- New tests: NAD listing (sorted join, namespace scoping, empty/missing cases), RoCE/IB/no-interconnect annotation gating, GID-index omission on both fabrics

### DIFF
--- a/cluster/kube/builder/builder.go
+++ b/cluster/kube/builder/builder.go
@@ -96,10 +96,19 @@ const (
 	// NCCL tuning vars, injected only for services whose reservation
 	// pinned an interconnect HCA. Tenants may override any of these by
 	// setting the same key in service.env (addIfNotPresent honors that).
-	envVarNCCLIBDisable     = "NCCL_IB_DISABLE"
-	envVarNCCLIBHCA         = "NCCL_IB_HCA"
-	envVarNCCLIBGIDIndex    = "NCCL_IB_GID_INDEX"
-	nccLIBGIDIndexRoCEValue = "3"
+	envVarNCCLIBDisable  = "NCCL_IB_DISABLE"
+	envVarNCCLIBHCA      = "NCCL_IB_HCA"
+	envVarNCCLIBGIDIndex = "NCCL_IB_GID_INDEX"
+)
+
+const (
+	// InterconnectFabricRoCE matches SchedulerResourceInterconnect.Fabric as
+	// stamped by tryAdjustInterconnect from the node's discovered link layer.
+	InterconnectFabricRoCE = "roce"
+
+	// multusNetworksAnnotation is the standard CNCF annotation multus watches
+	// to attach secondary networks (NetworkAttachmentDefinitions) to a pod.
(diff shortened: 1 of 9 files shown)


chalabi2 (approved):
lgtm!

troian (approved):

PR #429, opened 2026-08-05
feat: ngf http options via snippet filters

cloud-j-luna:
Providers to create `SnippetFilter` for options on Gateway providers

## Test evidence

```
1. All filters emitted and reaching the data plane. The provider's SnippetsFilter and the rendered nginx include are identical, and nginx -t accepts them:

$ nginx -t
nginx: configuration file /etc/nginx/nginx.conf test is successful

$ cat .../SnippetsFilter_http.server.location_<ns>_hello.localhost.conf
client_max_body_size 2097152;
proxy_read_timeout 60000ms;
proxy_send_timeout 60000ms;
proxy_buffer_size 16k;
proxy_buffers 8 16k;
proxy_next_upstream_tries 3;
proxy_next_upstream off;

2. Programmed and resolved (no dangling reference).

HTTPRoute hello.localhost:  Accepted=True  ResolvedRefs=True
SnippetsFilter hello.localhost: Accepted=True
Gateway akash-gateway:  Accepted=True  Programmed=True

3. client_max_body_size enforced end to end (curl through the gateway).

[GET /]                 -> 200     # route serves
[POST 3MB > 2MB limit]  -> 413     # nginx rejects, per client_max_body_size
[POST 1MB < 2MB limit]  -> 405     # under limit: passed nginx to the backend (405 = app rejects POST)

nginx confirms the cause of the 413:

[error] client intended to send too large body: 3145728 bytes, server: hello.localhost, request: "POST / HTTP/1.1"
```

### DIFF
--- a/cluster/kube/gateway/httproute.go
+++ b/cluster/kube/gateway/httproute.go
@@ -3,14 +3,19 @@ package gateway
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/pager"
+	"k8s.io/client-go/util/retry"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/akash-network/provider/cluster/kube/builder"
@@ -49,6 +54,14 @@ func (NoopHTTPRouteObserver) OnDelete(error) {}
 
 // CreateOrUpdateHTTPRoute creates or updates an HTTPRoute for a hostname directive.
 // It uses the provided Provider to build annotations and the HTTPRoute spec.
+//
+// Route extensions (an NGF SnippetsFilter) are applied before the HTTPRoute
+// publishes its ExtensionRef to them. A route that references a missing
+// SnippetsFilter makes NGF set ResolvedRefs=False and serve HTTP 500 for that
+// rule, so an existing route keeps its working spec until the filter is applied.
+// The SnippetsFilter is owner-referenced to the route for garbage collection, so
+// a brand-new route is created without the filter first to mint a UID, then the
+// filter is applied and only then is the reference added.
 func CreateOrUpdateHTTPRoute(
 	ctx context.Context,
 	dc dynamic.Interface,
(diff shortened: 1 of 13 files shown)


cluster/kube/gateway/httproute.go
```diff
@@ -73,40 +84,139 @@ func CreateOrUpdateHTTPRoute(
...
+	// spec, re-reading the resourceVersion on each attempt: NGF writes route status
+	// between our earlier reads and this update, so a cached resourceVersion can be
+	// stale and 409, which would otherwise leave the route without its filter.
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		current, gErr := routes.Get(ctx, routeName, metav1.GetOptions{})
+		if gErr != nil {
+			return gErr
+		}
+		u.SetResourceVersion(current.GetResourceVersion())
+		_, uErr := routes.Update(ctx, u, metav1.UpdateOptions{})
```
chalabi2, line 168: blocking: applyRouteExtensions only confirms that the Kubernetes API accepted the SnippetsFilter write not that NGF has observed it.

Since NGF watches HTTPRoutes and SnippetsFilters independently, this update can publish the ExtensionRef first, causing ResolvedRefs=False and HTTP 500 until the filter event catches up. Eventual consistency makes it recover, but it doesn't remove that failure window.

Can we either:

wait for Accepted=True at the written SnippetsFilter generation before publishing the reference, or
publish the reference on a later reconcile?

chalabi2 (changes requested):

cluster/kube/gateway/httproute.go
```diff
@@ -73,40 +86,202 @@ func CreateOrUpdateHTTPRoute(
...
+	}
+	routeExists := getErr == nil
+
+	var ownerUID types.UID
+	createdBare := false
+	if routeExists {
+		ownerUID = existing.GetUID()
+	} else if len(exts) > 0 {
+		bareSpec := spec
+		bareSpec.Rules = rulesWithoutFilters(spec.Rules)
```
chalabi2: Blocking: keep the placeholder HTTPRoute detached from the Gateway.

Removing only Filters leaves the route's ParentRefs, hostname, and backend intact, so this is already a publicly routable route. If filter creation, acceptance, or the final update fails, reconciliation returns while the backend remains exposed without the requested body-size, timeout, retry, or buffer settings.

I reproduced this against NGF 2.4.1: after a controller outage, the leftover route reached Accepted=True and ResolvedRefs=True, returned HTTP 200, and allowed a 2 KiB request through despite a requested 1 KiB limit.

Please create the placeholder without ParentRefs or routable rules, then attach the complete spec only after the SnippetsFilter is accepted. This also needs a new-route failure regression test.

cluster/kube/gateway/nginx_gateway.go
```diff
@@ -92,82 +110,123 @@ func (n *nginxGateway) BuildHTTPRouteSpec(
...
+		line("proxy_read_timeout %dms;", d.ReadTimeout)
+	}
+	if d.SendTimeout > 0 {
+		line("proxy_send_timeout %dms;", d.SendTimeout)
+	}
+	if proxyBufferSize != "" {
+		// proxy_buffer_size alone makes nginx's default proxy_busy_buffers_size
+		// (2x) exceed the default proxy_buffers pool, so the config is rejected on
+		// reload. Pair it with a matching proxy_buffers so it is valid for any size.
+		line("proxy_buffer_size %s;", proxyBufferSize)
```
chalabi2, line 176: Validate proxy-buffer-size before inserting it into raw NGINX configuration.

NGF marks a structurally valid SnippetsFilter as Accepted=True without parsing its NGINX contents. A plausible operator typo such as 16kb therefore passes the new readiness gate, but nginx -t rejects it with proxy_buffer_size directive invalid value, preventing the data plane from applying the configuration.

Please validate this flag at the CLI/config boundary and fail startup unless it is empty or a valid positive NGINX size.

chalabi2 (changes requested):
One additional blocking proof gap: the green E2E check does not run TestGatewayAPISuite. The integration target explicitly uses -run TestIntegrationTestSuite, and the GitHub Actions log shows that exact command. The new Gateway API test therefore compiles but never executes.

Please add a Gateway/NGF CI job that runs TestGatewayAPISuite and verifies SnippetsFilter Accepted=True, HTTPRoute ResolvedRefs=True, Gateway/data-plane programming, and at least one real HTTP option such as the body-size limit.

chalabi2 (approved):
lgtm

troian (approved):

PR #432, opened 2026-08-25
fix: close reclaiming leases once the reclamation window elapses

cloud-j-luna:
## Problem

A lease that enters the on-chain `reclaiming` state can stay there forever.
We hit this on mainnet (dseq `1781208244926`): the deployment showed as reclaimed but the workload kept running and the lease never closed on chain.

Root cause: after `MsgLeaseStartReclaim` moves a lease to `reclaiming` with a deadline, something has to broadcast `MsgCloseBid` once the reclamation window elapses.
The chain does not do it (the market `EndBlocker` is a no-op), and the provider had no path that did it either.
`cluster/service.go` only logged `EventLeaseReclaimStarted` and waited for an `EventLeaseClosed` that never arrives.
Disabling reclamation on the provider was a red herring: nothing closed the lease whether the offer flag was on or off.

## Fix

The provider already owns each of its leases through a per-lease `deploymentManager`, and already receives `EventLeaseReclaimStarted` (carrying the deadline) on its event bus.
This wires the close into that existing machinery rather than adding a separate reconciler.

When a lease enters reclamation, the manager that owns it arms a timer for the deadline.
Two paths feed the same seam, a `reclaim(deadline)` method on the manager that mirrors `update`/`teardown`:

- Live. `cluster/service.go` routes `EventLeaseReclaimStarted` to `s.managers[id]`.
- Restart recovery. On startup the manager is rebuilt from the still-running workload, and `checkLeaseActive` reads the reclaiming deadline off chain and arms the same timer.

When the timer fires the manager broadcasts `MsgCloseBid` (reason `Decommissioned`), and the resulting `EventLeaseClosed` drives the existing teardown.
The close is gated on the node's latest block time rather than wall clock, since the chain enforces the deadline against block time; this avoids broadcasting early closes the chain would reject and still charge fees for.
A not-yet-due or rejected close re-arms and retries.

Known gap: a lease reclaiming on chain with no local manager (an unusual desync where the workload is already gone) is cleaned on restart.
During a normal reclamation window the workload is still running, so the manager is present.


reclamation_closer.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+		}
+	}
+}
+
+// sweep closes every lease of this provider whose reclamation window has elapsed.
+func (rc *reclamationCloser) sweep(ctx context.Context) error {
+	sctx, cancel := context.WithTimeout(ctx, reclamationRPCTimeout)
+	syncInfo, err := rc.session.Client().Node().SyncInfo(sctx)
+	cancel()
```
troian: put `cancel` into defer

reclamation_closer.go
```diff
@@ -0,0 +1,158 @@
...
+	if syncInfo.CatchingUp {
+		return nil
+	}
+
+	provider := rc.session.Provider().Owner
+
+	var nextKey []byte
+	for {
+		qctx, cancel := context.WithTimeout(ctx, reclamationRPCTimeout)
+		resp, err := rc.aqc.Market().Leases(qctx, &mvbeta.QueryLeasesRequest{
```
troian: there is lease querier at provider startup, we don't have to duplicate it

troian (changes requested):
how does `closer` know lease entered reclamation stage

cluster/service.go
```diff
@@ -384,8 +384,16 @@ loop:
...
+				// Workloads keep running during the reclamation window. The lease's manager
+				// arms a timer for the deadline and broadcasts MsgCloseBid once it elapses in
+				// block time; the resulting EventLeaseClosed drives teardown here.
+				if manager := s.managers[ev.ID]; manager != nil {
+					if err := manager.reclaim(ev.Deadline); err != nil {
+						s.log.Error("arming reclamation close", "err", err, "lease", ev.ID)
+					}
+				} else {
+					s.log.Warn("reclaim event for lease with no local manager", "lease", ev.ID)
+				}
```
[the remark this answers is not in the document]
troian: update error should not lead to manager exit

cloud-j-luna:
Refactored from polling based `closer` to event-driven. This plugs into the existing event mechanisms

cluster/service.go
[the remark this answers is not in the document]
cloud-j-luna: Addressed as on the prior deploy's error we now log and fall through to startDeploy(ctx), which applies the queued update instead of tearing the manager down

cloud-j-luna:
Retriggering CI

cloud-j-luna:
Pending stuck checks, retriggering CI

cluster/manager.go
```diff
@@ -87,6 +103,7 @@ func newDeploymentManager(s *service, deployment ctypes.IDeployment, isNewLease
 		wg:                  sync.WaitGroup{},
 		updatech:            make(chan ctypes.IDeployment),
 		teardownch:          make(chan struct{}),
+		reclaimch:           make(chan int64, 1),
```
troian, line 101: `chan time.Duration`

cloud-j-luna: I see what you meant now, this is int64 because its the absolute deadline to reclaim, duration is used for the wait/retry interval

cloud-j-luna:
Both `reclamationRPCTimeout`/`reclamationCloseRetryInterval` are now configurable

PR #433, opened 2026-09-03
Feat/cpu arch inventory

chainzero:


troian (approved):

chainzero:
Closing the loop on the automated review findings (claude[bot] + coderabbitai[bot]) now that this has merged and been validated in sandbox:

1. Default-to-amd64 when arch is unspecified (inventory.go:405)
Intentional, not a bug. Defaulting unspecified arch to amd64 leaves the existing all-amd64 fleet completely unaffected (every current provider matches trivially) and requires arm64 to be an explicit tenant opt-in — protecting tenants from being silently scheduled onto incompatible hardware. No change.

2. Missing CPU NodeSelector/affinity on the Pod spec (workload.go / inventory.go:407)
Per the AEP, v1 of arm64 support intentionally targets homogeneous clusters only — a provider is either all-amd64 or all-arm64, never mixed. This gap only matters when a single cluster has more than one architecture present, so it's explicitly out of scope for this release rather than a hole in it. Heterogeneous-cluster support (and this fix, mirroring the existing GPU NodeSelector pattern) would be a distinct future decision if that scope ever changes.

3. Probe-failure ambiguity — empty CPUInfo after retries falls back to "legacy/compatible" (node-discovery.go:1202)
Real, and not fully ruled out by #2's homogeneous-only scoping — this one is node-local (a single node's persistent sidecar failure), not cluster-topology-dependent. Requires a sustained probe failure surviving the full retry budget, which is uncommon, combined with a tenant explicitly requesting the architecture that node doesn't actually have. Not fixing now given the rarity/complexity tradeoff (the correct fix needs a new "unknown vs. legacy" state threaded through inventory types); tracking as a follow-up.

4. Retry-loop doesn't exit early on context cancellation (node-discovery.go:1187)
Confirmed low severity on inspection: the loop is hard-capped at 5 attempts regardless of this fix. Correcting it would only mean exiting early and saving some of those already-capped attempts on cancellation — the end result (empty CPUInfo, same error logged) is identical either way, no behavioral or data consequence. Tracking as a minor follow-up fix given it's cheap and correctly identified, not urgent.

PR #284, opened 2025-04-25
chore: add capability of e2e tests to observe provider cluster state and check workload side effects

cloud-j-luna:
This PR adds the capability of observing a provider cluster state and performing verifications.
It adds several e2e tests to verify if workloads had unexpected impact from several scenarios.

## Current tests

- [x] Deployment upgrade creates a new pod.
- [x] Deployment shell access doesn't trigger workload restarts.
- [x] Provider & Hostname Operator restarts don't impact deployed workloads.

PR #434, opened 2026-09-08
fix(kube): preserve workloads across provider recovery

chalabi2:
Since #419, recovery replaces unchanged tenant pods: rewriting the saved Manifest's resource-version label bumps the Manifest's own resource version, and a later recovery copies that into pod templates, triggering a rollout. Repro passes at #419's parent, fails at #419 and main.

Fix: preserve Manifest metadata, keep version bookkeeping on workload objects. Existing pod-template labels keep their values so upgrades don't rollout. Tenant updates still apply.

Tests: every deployment fixture now runs repeated recovery, unchanged resend, tenant update, recovery-after-update across the full feature matrix. CI also snapshots workloads with base code and recovers them with the candidate, catching new defaults same-version tests miss. Matrix uses fake clients, so no real RDMA/GPU/CC hardware.

Refs #419. Supersedes #284, which only tested one restart; this regression needs two.

The full-history checkout preserves annotated tag refs so signed release tags can be fetched during CI setup. The alpha tag run exposed this checkout failure; the corrected tag run passes the full unit suite and release dry-run.


chalabi2:
@claude review

troian (dismissed):

troian (approved):

PR #437, opened 2026-09-23
fix(cluster): tear down leases closed on chain when the close event is missed

cloud-j-luna:
## Problem and root cause

A lease that is closed on chain can keep running on the provider indefinitely, holding its Kubernetes workload and reserved resources (for GPU leases the GPUs stay locked), while the balance-checker retries withdrawing from the already-settled escrow and logs `payment closed ... unknown request` every ~10 minutes. Observed on mainnet (dseq `28684233`): deployment, group, order, bid and lease all `closed` on chain with escrow settled to zero, yet the workload kept running. The provider tears a lease down only when it receives `EventLeaseClosed` on its pubsub bus (`cluster/service.go`, which also stops the balance-checker via `LeaseRemoveFundsMonitor`), and that event is fed by the chain-event subscription (`pkg.akt.dev/go/util/events`) over the CometBFT websocket. When that subscription drops and cannot recover (the ws client reconnects with an uncapped `(1 << attempt)`-second backoff and, after `maxReconnectAttempts`, stops entirely), the provider silently stops receiving all chain events, so `EventLeaseClosed` never arrives. Nothing self-heals while the provider stays up, because the only manager-versus-chain reconcile runs solely at manager creation, which is why a restart is the current workaround.

## Fix

The balance-checker already polls every lease it monitors on a timer, and its `doEscrowCheck` already fetches the deployment and escrow account to compute remaining funds. That same response already tells us the deployment is closed. This change reads it: when the fetched deployment is `DeploymentClosed`, the balance-checker publishes `EventLeaseClosed` itself, so the existing teardown path runs exactly as it would for a real chain event (workload teardown plus `LeaseRemoveFundsMonitor`, which also stops the balance-checker). The cost is zero additional RPC, since it reuses a query the balance-checker already makes, and it lives in the component that was otherwise spinning on the closed lease. It is conservative: it fires only on an explicit `DeploymentClosed` state, never on a query error or a not-yet-closed deployment, so a lagging RPC backend cannot force a false teardown; and if the publish fails it retries on the next check rather than latching. This supersedes an earlier per-`deploymentManager` polling approach that issued one extra lease query per lease per interval (O(N) added RPC, more load on the very proxy whose flakiness caused the bug); the piggyback removes that machinery entirely. It is complementary to #432, which handles leases stuck in the on-chain `reclaiming` state where the chain never emits `EventLeaseClosed`. Verification: `go build ./...`, `go test ./...`, and `golangci-lint` are green, and a new in-process test (`balance_checker_test.go`) wires a real balance-checker and pubsub bus to a mocked chain and asserts that a closed deployment produces `EventLeaseClosed` while an active one does not. Known scope: detection is at deployment granularity, so a single closed lease within a still-active multi-group deployment is not covered by this backstop; that case still relies on the normal event path.


cloud-j-luna:
Superseded by akash-network/chain-sdk#361: the fix moved to the chain-sdk events service (self-healing block polling instead of a websocket subscription that could silently die). The provider side is just a dependency bump once that releases, no code change.

